PR #1, opened 2018-11-01
Fix ARM_GICD_IROUTER constant to match the ARM specification

chris-oo:
SPIs start at 32, but the constant ARM_GICD_ROUTER skips this by defining the start as 0x6100 vs what the specification says, 0x6000. This is unintuitive and could lead to bugs due to needing to subtract by 32 whenever programming SPI IROUTERs.


corthon:
Is it your intention to only submit this to 180529?
1) Is it breaking things in a way that justifies pushing this into an ostensibly closed branch?
2) Is it also needed in 1808?

spbrogan:
Adding link for tianocore bug.  
https://bugzilla.tianocore.org/show_bug.cgi?id=1294

corthon:
What's up with this PR? Is it here for public comment? Still needed in the code?
@spbrogan , @chris-oo

PR #2, opened 2018-12-18
matched parameters for LcdInitialize

matthewfcarlson:
It was causing ArmMaliDp to fail to get built, which threw the build process off track


corthon:
@spbrogan and @dmitryantipov,
Was this PR gate ever passing? Doesn't look like Matt's changes had anything to do with this.

corthon:
If this change is still necessary/valid, we need to restage it against 201903. Sorry for the delay and frustration.

PR #13, opened 2021-11-04
Remove define ID_AA64MMFR2_EL1 in AArch64Support.masm, because it is now present in VS2019 toolchain

mebersol:
Remove define for ID_AA64MMFR2_EL1

### DIFF
--- a/ArmPkg/Library/ArmLib/AArch64/AArch64Support.masm
+++ b/ArmPkg/Library/ArmLib/AArch64/AArch64Support.masm
@@ -72,8 +72,6 @@
 #define CTRL_V_BIT       (1 << 12)
 #define CPACR_VFP_BITS   (3 << 20)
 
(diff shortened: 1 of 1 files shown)


corthon (approved):
Do we have any justification for this? If so, can you add it in the commit notes? The behavior itself is pretty self-evident, but any reviewer in the future would want to know why it was there before and why it's no longer needed now.

spbrogan (approved):

spbrogan (commented):
Should we wrap this in a VS2019 version check?

mebersol:
> Should we wrap this in a VS2019 version check?
What would this look like?

spbrogan:
I need to look up how to do it in the assembler. But here is an example for C code. 

https://github.com/microsoft/mu_basecore/blob/release/202102/MdePkg/Include/X64/ProcessorBind.h#L104

PR #20, opened 2022-10-24
Change the package inclusion over to QEMU package

kuqin12:


ArmVirtPkg/ArmVirtPkg.ci.yaml
```diff
@@ -56,7 +56,8 @@
             "ArmPlatformPkg/ArmPlatformPkg.dec",
             "PcAtChipsetPkg/PcAtChipsetPkg.dec",
             "SecurityPkg/SecurityPkg.dec",
-            "ShellPkg/ShellPkg.dec"  #Is this ok?
+            "ShellPkg/ShellPkg.dec",  #Is this ok?
```
spbrogan, line 59: can you scope this to UEFI APPLICATION or does your firmware code really need this?

ArmVirtPkg/ArmVirtPkg.ci.yaml
```diff
@@ -56,7 +56,8 @@
             "ArmPlatformPkg/ArmPlatformPkg.dec",
             "PcAtChipsetPkg/PcAtChipsetPkg.dec",
             "SecurityPkg/SecurityPkg.dec",
-            "ShellPkg/ShellPkg.dec"  #Is this ok?
+            "ShellPkg/ShellPkg.dec",  #Is this ok?
+            "QemuQ35Pkg/QemuQ35Pkg.dec"  #This probably not ok?
```
spbrogan, line 60: i think this points at the need for common virtual platform package.

ArmVirtPkg/ArmVirtPkg.ci.yaml
kuqin12: This is needed in the `PlatformHasAcpiDtDxe.inf` DXE driver I pulled in for ARM platform. This driver is using a library header file originally defined in OvmfPkg..

ArmVirtPkg/ArmVirtPkg.ci.yaml
kuqin12: That seems right.. But it will eventually be as rich as the OvmfPkg. Also I think this package's hierarchy is off.

kuqin12:
Turned out that this driver is not needed for the platform we are trying to onboard. Will leave this one closed till later when we need it again.

PR #34, opened 2022-12-07
changes to support setting OSC CTRL bits based on PCI config space info struct

JoeLopez333:
## Description

The OSC method in the context of PCI conveys information to the Operating System on which features it may take control of (native control) and which features will be controlled by firmware (firmware-first).

The previous version of this library uses a hard-coded OSC method that does not allow the consumer to specify if certain features should be granted to the OS or not. 

This change uses the aml fixup libraries to set the CTRL bits in the OSC method.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Tested on platform that utilized DyamicTablesPkg. Verified that selected osc template will be used in the PCI ACPI device when specified in the PCI Config Space Info structure
## Integration Instructions
Consumers who utlize the DynamicTablesPkg should specify in CM_ARM_PCI_CONFIG_SPACE_INFO.OscControlBitmap which OSC CTRL bits should be set if the OS requests them


DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -420,6 +420,19 @@ typedef struct CmArmGenericWatchdogInfo {
   UINT32    Flags;
 } CM_ARM_GENERIC_WATCHDOG_INFO;
 
+/** The ARM_PCI_OSC_INFO struct describes the features that FW will
+ * grant native control over via _OSC method.
+*/
```
os-d: nit: add a space in front of ending comment symbol

JoeLopez333:
@os-d @kuqin12 These updates should reflect the changes you suggested, looking forward to your feedback!

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
```diff
@@ -46,21 +46,16 @@ DefinitionBlock ("SsdtPciOsc.aml", "SSDT", 2, "ARMLTD", "PCI-OSC", 1) {
...
-      Store (CDW3, CTRL)
+      And (CTRL, CDW3, CTRL)
 
       // Only allow native hot plug control if OS supports:
       // * ASPM
       // * Clock PM
       // * MSI/MSI-X
       If (LNotEqual (And (SUPP, 0x16), 0x16)) {
-        And (CTRL, 0x1E, CTRL) // Mask bit 0 (and undefined bits)
+        And (CTRL, 0x9E, CTRL) // Mask bit 0 (and undefined bits)
```
os-d: Can you explain the switch to mask `9E` here? Is this keeping native hot plug control if the OS doesn't support the required features? Perhaps I am reading my bits incorrectly :)

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -420,6 +420,48 @@ typedef struct CmArmGenericWatchdogInfo {
...
+    /// Allow Native Completion Timeout Control
+    UINT32     AllowNativeCompletionTimeoutControl : 1;
+
+    /// Allow Native System Firmware Intermediary Configuration Control
+    UINT32     AllowNativeSfiControl : 1;
+
+    UINT32     Reserved22 : 22;
+  };
+  UINT32     AsUint32;
+} ARM_PCI_OSC_CONTROL_BUFFER_INFO;
```
kuqin12: Can we please add `MU_CHANGE` tags around the updated code blocks? It will help the integration process next time we pick up the next release tag. Thanks!

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
```diff
@@ -46,21 +46,16 @@ DefinitionBlock ("SsdtPciOsc.aml", "SSDT", 2, "ARMLTD", "PCI-OSC", 1) {
...
       }
 
-      // Always allow native PME, AER (no dependencies)
-
-      // Never allow SHPC (no SHPC controller in this system)
```
kuqin12, line 61: Same here, can we please add a `MU_CHANGE` tag and just comment out this section?

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
kuqin12: Dumb question: what is this extra bit for? Do we need to update the comment?

kuqin12 (approved):

os-d (approved):

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
os-d: Chatted offline, I was reading my bits incorrectly. This masks off native hot plug control if the OS doesn't support it. With the new capability of setting more bits, we will change this to leave the other defined bits and only mask off native hotplug support.

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
```diff
@@ -46,21 +46,16 @@ DefinitionBlock ("SsdtPciOsc.aml", "SSDT", 2, "ARMLTD", "PCI-OSC", 1) {
 
       // Save Capabilities DWord2 & 3
       Store (CDW2, SUPP)
-      Store (CDW3, CTRL)
+      And (CTRL, CDW3, CTRL)
```
pierregondois, line 52: Hello,
I think this statement has no effect:
CTRL is initialized to 0, so
`And (CTRL, CDW3, CTRL)`
should do:
`CTRL = 0 & CDW3`
All the following 'and' operations on CTRL might have no effect aswell

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
pierregondois: Hi,
Just for information, this piece of code comes from the 'PCI Firmware Specification Revision 3.3', s4.5.3. 'ASL Example'.

It seems that in revision 3.0, only bits [0:4] were defined (cf Table 4-5 and 4-6: Interpretation of the _OSC Control Field), and the example only saves these bits (except bit 0). Now that there are more bits defined, the example in the spec (and the original code in this file) might be out-dated.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -420,6 +420,48 @@ typedef struct CmArmGenericWatchdogInfo {
...
+
+    /// Allow Native Completion Timeout Control
+    UINT32     AllowNativeCompletionTimeoutControl : 1;
+
+    /// Allow Native System Firmware Intermediary Configuration Control
+    UINT32     AllowNativeSfiControl : 1;
+
+    UINT32     Reserved22 : 22;
+  };
+  UINT32     AsUint32;
```
pierregondois: Hi,
I never had this issue, but I thought it was not advised to use bit-fields like that as it is not guaranteed that each field will conserve its position (i.e. AllowNativeHotPlugControl might not be mapped to the first bit of the resulting uint32).
I usually see bit-fields used as masks:
`#define PIC_OSC_ALLOW_NATIVE_HOT_PLUG_CONTROL      0x01`

os-d: This is fair, I think in practice this will work, but the C programming language book does say: "Almost everything about fields is implementation-dependent." It is likely safer to define the bitfields as Pierre showed.

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
os-d: My understanding is that the added code in SsdtPcieSupportLib will fixup the 0 defined in the asl code here to be the value of the `ARM_PCI_OSC_CONTROL_BUFFER_INFO` union which the platform can now define which bits are set in that. Does this clash with your understanding of fixing up the asl named object?

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
os-d: Thanks Pierre, this is a good call out. In revisiting the spec, I agree, this is an example for a limited mobile system, in order to genericize this, we need to ensure we are looking at the expanded bit requirements (like DPC for example) and enforcing them.

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
JoeLopez333: Oliver is correct, the CTRL Nameop is fixed up in SsdtPcieSupportLib according to the bits in CM_ARM_PCI_CONFIG_SPACE_INFO.OscControlBuffer, I have verified that the fixup works as expected to replace the 0 value and updated the comment to reflect that the Nameop is a template which will be overridden later

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
JoeLopez333: Added MU_CHANGE tags to all added code, thanks Kun

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
JoeLopez333: updated comment format to align with other comments in this file

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
JoeLopez333: Thanks Pierre, Oliver.

I have update the asl code to reflect the dependencies outlined in the FW spec v 3.3

JoeLopez333: The extra bit (bit 7) is for granting native DPC control to the OS

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
JoeLopez333: Thanks for the note, updated to use #define masks instead of a struct/union

DynamicTablesPkg/Library/Common/SsdtPcieSupportLib/SsdtPcieOscTemplate.asl
pierregondois: Ok yes right, I didn't see the `CTRL` value was modified

PR #35, opened 2022-12-08
[Rebase & FF] Introducing new PSCI command index and update SGI function

kuqin12:
## Description

This change added a few minor upgrades for existing ArmPkg assets:
- Added assembly function to write to `ICC_SGI1R` from non-secure EL1/EL2
- Extended current `ArmGicSendSgiTo` interface to also support sending SGI on GIC v3 and v4
- Added a new PSCI command index to query PSCI suspend features.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on both QEMU ARM platforms and FVP.

## Integration Instructions

Extended functionality. No existing platforms integration steps needed.


ArmPkg/Drivers/ArmGic/ArmGicLib.c
```diff
@@ -139,10 +139,53 @@ ArmGicSendSgiTo (
   IN  INTN  SgiId
   )
 {
-  MmioWrite32 (
-    GicDistributorBase + ARM_GIC_ICDSGIR,
-    (UINT32)(((TargetListFilter & 0x3) << 24) | ((CPUTargetList & 0xFF) << 16) | SgiId)
-    );      // MS_CHANGE
+  // MU_CHANGE Starts: Added check for supported GIC version and expanded support for sending SGI on GICv3 and GICv4.
```
os-d, line 142: The comments say GICv3 and v4, but the code only calls a GICv3 API. Is it the same API for GICv4?

os-d (approved):

ArmPkg/Drivers/ArmGic/ArmGicLib.c
kuqin12: Yeah, the implementation for this function is essentially following the doc: https://developer.arm.com/documentation/198123/0302/Sending-and-receiving-Software-Generated-Interrupts (I did not add it since I am afraid it may disappear someday). But the `ICC_SGI1R_EL1` function was following the existing semantics of existing ArmGicLib, which only has distinction over v2 and v3.

cfernald (approved):

PR #37, opened 2022-12-16
Specify function number in PCI dynamic ACPI object

JoeLopez333:
## Description

Allow user to specify Function number in InterruptInfo structure

In original implementation, assumption was made that function number should be 0xFFFF in _PRT and DXX_ objects. This implies that the asl object applies to all functions under a PCI device.

It is a valid use case to specify only certain function numbers under a device in these objects, on certain operating systems it is even necessary to not use 0xFFFF and instead specify a specific function number.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Tested on platform that utilizes DynamicTablesPkg

## Integration Instructions

Ensure that desired function number is added in the Interrupt Info structure.


kuqin12 (approved):

os-d (approved):
Will you update the PR title to reflect the change made?

PR #63, opened 2023-02-14
Create ARM platform library function for customizing PEI memory

cfernald:
## Description

Creates a new function in the ARM platform library to allow the platform to customize the location of the initial PEI memory installation region.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested against the SBSA platform with custom memory initialization logic.

## Integration Instructions

Platforms with their own arm platform package will need to add implementation of ArmPlatformGetPeiMemory that returns FALSE to be unaffected.


ArmPlatformPkg/Include/Library/ArmPlatformLib.h
```diff
@@ -136,4 +136,20 @@ ArmPlatformGetPlatformPpiList (
...
+
+  @param[out]  EfiMemoryBase  The custom memory base, should be unchanged if FALSE is returned.
+  @param[out]  EfiMemorySize  The custom memory size, should be unchanged if FALSE is returned.
+
+  @retval   TRUE    A custom memory region was set.
+  @retval   TRUE    A custom memory region was not set.
+**/
+BOOLEAN
+EFIAPI
+ArmPlatformGetPeiMemory(
```
kuqin12: nit: MU_CHANGE tags?

kuqin12 (approved):

ArmPlatformPkg/Include/Library/ArmPlatformLib.h
```diff
@@ -136,4 +136,24 @@ ArmPlatformGetPlatformPpiList (
   OUT EFI_PEI_PPI_DESCRIPTOR  **PpiList
   );
 
+// MU_CHANGE START: Allow platform to customize initial memory region.
+
+/**
+  Checks if the platform requires a special initial EFI memory region.
+
+  @param[out]  EfiMemoryBase  The custom memory base, should be unchanged if FALSE is returned.
+  @param[out]  EfiMemorySize  The custom memory size, should be unchanged if FALSE is returned.
```
os-d: nit: should be -> will be (I like strong contracts :) if that isn't true, then I would call that out specifically.

ArmPlatformPkg/Include/Library/ArmPlatformLib.h
```diff
@@ -136,4 +136,24 @@ ArmPlatformGetPlatformPpiList (
...
+// MU_CHANGE START: Allow platform to customize initial memory region.
+
+/**
+  Checks if the platform requires a special initial EFI memory region.
+
+  @param[out]  EfiMemoryBase  The custom memory base, should be unchanged if FALSE is returned.
+  @param[out]  EfiMemorySize  The custom memory size, should be unchanged if FALSE is returned.
+
+  @retval   TRUE    A custom memory region was set.
+  @retval   TRUE    A custom memory region was not set.
```
os-d: should be FALSE here

ArmPlatformPkg/Include/Library/ArmPlatformLib.h
os-d: Same below.

ArmPlatformPkg/Include/Library/ArmPlatformLib.h
os-d: Same below

ArmPlatformPkg/MemoryInitPei/MemoryInitPeim.c
```diff
@@ -114,8 +115,16 @@ InitializeMemory (
...
   //
 
-  // In case the firmware has been shadowed in the System Memory
-  if ((FdBase >= SystemMemoryBase) && (FdTop <= SystemMemoryTop)) {
+  // MU_CHANGE START: Allow platform to customize initial memory region.
+  UefiMemorySize = FixedPcdGet32 (PcdSystemMemoryUefiRegionSize);
+  if (ArmPlatformGetPeiMemory (&UefiMemoryBase, &UefiMemorySize)) {
+    // Check the Firmware does not intersect with the provided memory region.
+    ASSERT ((FdBase < UefiMemoryBase) || (FdBase >= (UefiMemoryBase + UefiMemorySize)));
+    ASSERT ((FdTop <= UefiMemoryBase) || (FdTop > (UefiMemoryBase + UefiMemorySize)));
```
os-d, line 123: in this case should we do anything besides ASSERT? i.e. in the release case is falling through the desired behavior or should we revert to the old behavior?

os-d (approved):

ArmPlatformPkg/MemoryInitPei/MemoryInitPeim.c
cfernald: I don't think we should fall back to the old behavior because the reason for the customization might have significant issues if it is not followed depending on the scenario. To me this falls under the same class of issue that exists everywhere in PEI where allocations and memory layout needs to be done with good design as overlapping allocations/usages are not strictly checked in general.

PR #200, opened 2024-01-24
Implement Memory Attribute Protocol Installation Policy Option

TaylorBeebe:
## Description

Linux shim currently incorrectly uses the UEFI memory attribute protocol
causing a fault. The broken shim does not have the NXCOMPAT
flag, so compatibility mode can be used to uninstall the protocol
when it is loaded. For flexibility, this patch implements the policy
configuration option to allow ARM platforms to choose not to install the
protocol.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on SBSA by running the DxePagingAuditTestApp with the protocol policy set to TRUE and FALSE

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/CpuDxe.c
+++ b/ArmPkg/Drivers/CpuDxe/CpuDxe.c
@@ -327,14 +327,26 @@ CpuDxeInitialize (
   //
   RemapUnusedMemoryNx ();
 
+  // MU_CHANGE [START]: Only install Memory Attribute Protocol if policy is enabled
   Status = gBS->InstallMultipleProtocolInterfaces (
                   &mCpuHandle,
                   &gEfiCpuArchProtocolGuid,
                   &mCpu,
-                  &gEfiMemoryAttributeProtocolGuid,
-                  &mMemoryAttribute,
+                  // &gEfiMemoryAttributeProtocolGuid,
+                  // &mMemoryAttribute,
                   NULL
                   );
+
+  if (gDxeMps.InstallMemoryAttributeProtocol) {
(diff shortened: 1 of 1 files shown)


ArmPkg/Drivers/CpuDxe/CpuDxe.c
```diff
@@ -327,15 +327,27 @@ CpuDxeInitialize (
...
                   );
 
+  if (gDxeMps.InstallMemoryAttributeProtocol) {
+    Status = gBS->InstallMultipleProtocolInterfaces (
+                    &mCpuHandle,
+                    &gEfiMemoryAttributeProtocolGuid,
+                    &mMemoryAttribute,
+                    NULL
+                    );
+  }
```
os-d, line 351: We already weren't checking Status after this call, but perhaps we should be in both cases?

os-d (approved):

ArmPkg/Drivers/CpuDxe/CpuDxe.c
os-d: In the first case, it would be fatal, in the second case, we could just log it (maybe assert, too, for debugging).

TaylorBeebe: Updated to assert on the first install, and print a DEBUG_ERROR message on the second

cfernald (approved):

PR #210, opened 2024-02-09
Remove print in AdvLogger path causing infinite recursion.

cfernald:
## Description

Removes a print that is in the AdvLogger initialization path that is causing infinite recursion as the print will reenter the AdvLogger initialization code. This code unfortunately cannot be made reentrant safe as at the time of calling the data sections are not guaranteed to be writable without first calling this routine, so the only clean solution is to remove this print.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested boot manually.

## Integration Instructions

N/A


kuqin12 (approved):

os-d (approved):

kenlautner (approved):

cfernald:
As discussed offline, this will be a temporary issue as we resolve more fundamental issues with the intersection of ARM MM and AdvLogger.

PR #220, opened 2024-03-14
Pre-Allocate Page Table Memory in ArmMmuLib

TaylorBeebe:
## Description

Allocating memory when memory protection is active can cause the below infinite loop:

1. MemoryAttributeProtocol->SetMemoryAttributes(EFI_MEMORY_RO)
2. ArmSetMemoryRegionReadOnly ()
3. SetMemoryRegionAttribute()
4. UpdateRegionMapping()
5. UpdateRegionMappingRecursive()
6. AllocatePages() -> Need memory for a translation table entry
7. CoreAllocatePages()
8. ApplyMemoryProtectionPolicy() -> Policy says new page should be XN
9. gCpu->SetMemoryAttributes()
10. Back to 3

To fix this previously, CpuDxe would update conventional memory to be XN prior to installing the CpuArch protocol. However, when we transition to setting EFI_MEMORY_RP on free memory, this will no longer work.

This PR updates ArmMmuLib to reserve page table memory for allocation during table spits to prevent the infinite loop.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on SBSA by creating the scenario where the infinite loop (without the XN remap routine in place) and booting successfully. This was also tested using the EFI_MEMORY_RP on free memory feature branch.

## Integration Instructions

Platforms which are using ArmMmuBaseLib for PEIM, PEI_CORE, and SEC modules will need to switch those module types to use ArmMmuPeiLib.
```
[LibraryClasses.common.SEC, LibraryClasses.common.PEIM, LibraryClasses.common.PEI_CORE]
  ArmMmuLib|ArmPkg/Library/ArmMmuLib/ArmMmuPeiLib.inf
```


ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -0,0 +1,121 @@
...
+
+  if (PoolPages == 0) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  //
+  // Avoid an infinite loop by bailing if we are nested in another call
+  // to this function
+  //
+  if (mPageTablePoolLock) {
```
os-d: Does this need to be a real lock? Is there a chance that we are preempted in this function by something that allocates memory?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -0,0 +1,121 @@
...
+    return EFI_ABORTED;
+  }
+
+  mPageTablePoolLock = TRUE;
+  PoolPages++;
+  PoolPages = ALIGN_VALUE (PoolPages, EFI_SIZE_TO_PAGES (SIZE_2MB)); // Add one page for the header
```
os-d: Where is the extra page?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -0,0 +1,121 @@
...
+  Buffer    = AllocateAlignedPages (PoolPages, BASE_2MB);
+  if (Buffer == NULL) {
+    DEBUG ((DEBUG_ERROR, "ERROR: Out of aligned pages\r\n"));
```
os-d: nit: drop `\r`, put function in here?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -0,0 +1,121 @@
...
+    goto Done;
+  }
+
+  if (mPageTablePool == NULL) {
+    mPageTablePool           = Buffer;
+    mPageTablePool->NextPool = mPageTablePool;
+  } else {
+    ((PAGE_TABLE_POOL *)Buffer)->NextPool = mPageTablePool->NextPool;
+    mPageTablePool->NextPool              = Buffer;
+    mPageTablePool                        = Buffer;
```
os-d: Perhaps I am misreading this, but doesn't this drop mPageTablePool itself? Can we use the standard list functions here to not reimplement?

os-d: If the goal is to drop mPageTablePool, should we make sure the existing pool is empty, first?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuPeiLibPageTableAlloc.c
```diff
@@ -0,0 +1,140 @@
...
+typedef struct {
+  VOID     *NextPool;
+  UINTN    Offset;
+  UINTN    FreePages;
+} PAGE_TABLE_POOL;
+
+#pragma pack()
+
+PAGE_TABLE_POOL  *mPageTablePool    = NULL;
+BOOLEAN          mPageTablePoolLock = FALSE;
```
os-d: We couldn't have globals for XIP PEI code, right, since the flash part is RO? While ARM64 doesn't have the same pre-mem constraints of x86, we still have early code executing XIP, although I need to recollect whether we rely on EL3 to load us into memory and so XIP is from memory and this is fine.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuPeiLibPageTableAlloc.c
```diff
@@ -0,0 +1,140 @@
...
+        break;
+      }
+
+      if (CurrentPool->NextPool == mPageTablePool) {
+        break;
+      }
+
+      CurrentPool = CurrentPool->NextPool;
+    }
+  }
```
os-d: I'm missing why we have the difference between PEI and DXE here, with PEI searching through the list and DXE not?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -274,7 +306,10 @@ UpdateRegionMappingRecursive (
         // No table entry exists yet, so we need to allocate a page table
         // for the next level.
         //
-        TranslationTable = AllocatePages (1);
+        // MU_CHANGE START: Use reserved page table memory pool
+        // TranslationTable = AllocatePages (1);
+        TranslationTable = AllocatePageTableMemory (1);
```
os-d, line 295: One question here: I know I advocated for a different approach in your other PR, but with the current approach of having memory attributes in ApplyMemoryProtectionPolicy, could we not just call AllocatePoolPages here, which skips the memory attribute application? I think preallocating is a good idea, but if we don't change the implementation of free memory RP, do we need it?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
TaylorBeebe: When we increment PoolPages we're adding an extra page, but we always allocate a size aligned to 2MB. So if a call is made to this to allocate 2MB, it will actually allocate 4MB and if a call is made to allocate 3MB it will also allocate 4MB.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
TaylorBeebe: Yes will drop -- this is a carryover from when I thought we used both

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
TaylorBeebe: I'm pretty closely mirroring the logic from CpuPageTable.c used for x86 platforms which I agree is not intuitive. Should I rewrite?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuPeiLibPageTableAlloc.c
TaylorBeebe: Pei should use ArmMmuPeiLib.inf which doesn't do any of this. I can remove the PEIM and PEI_CORE from the LIBRARY_CLASS section of ArmMmBaseLib.inf.

TaylorBeebe: Ah wait I see that this is the Pei version... This is a copy-paste issue when I was carrying things over from some tests in an Ubuntu container. Let me revert this back to its original.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
os-d: Oh, ha, I quite literally missed the increment directly above this.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
os-d: If it doesn't require backbending to use the standard list functions, I would prefer that, it is safer than having a bespoke implementation

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuPeiLibPageTableAlloc.c
os-d: Closing per comment above that PEI code will be removed.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
TaylorBeebe: I think CoreAllocatePoolPages is internal to DxeCore. Is it exposed somehow?

os-d: Ah, good point. This is non-core code, because what is core about manipulating the page table :)

TaylorBeebe: This library is both convenient and dangerous 😆

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuPeiLibPageTableAlloc.c
TaylorBeebe: Fixed. The PEI version of this does nothing in the initialize call and redirects to AllocatePages when AllocatePageTableMemory is called.

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
TaylorBeebe: Removed

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
TaylorBeebe: I think I may update the reserved page table pool to be fetched via protocol and owned by Arm's CpuDxe to make it a hybrid of the x86 approach and the Arm approach to page management. This will prevent a lot of strange edge cases which could occur if more than one module include ArmMmuLib and reduce the memory footprint.

TaylorBeebe: Updated to put the reserved pool in CpuDxe. I also updated the page table memory allocation behavior to keep the number of preallocated pages above 5.

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,193 @@
...
+/**
+  Allocate memory for the page table.
+
+  @param[in]  NumPages          Number of pages to allocate.
+
+  @retval A pointer to the allocated page table memory or NULL if the
+          allocation failed.
+**/
+VOID *
+AllocatePageTableMemory (
```
kuqin12, line 112: How would this mechanism prevent re-entrant calls if memory manager need to apply page table attributes in place and free page table buffer ran out?

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,193 @@
...
+
+  if (PoolPages == 0) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  PoolPages++;
+  PoolPages = ALIGN_VALUE (PoolPages, EFI_SIZE_TO_PAGES (SIZE_2MB)); // Add one page for the header
+  Buffer    = AllocateAlignedPages (PoolPages, BASE_2MB);
+  if (Buffer == NULL) {
+    DEBUG ((DEBUG_ERROR, "ERROR: Out of aligned pages\r\n"));
```
os-d: nit

```suggestion
    DEBUG ((DEBUG_ERROR, "ERROR: Out of aligned pages\n"));

```

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,193 @@
...
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Done;
+  }
+
+  if (mPageTablePool == NULL) {
+    mPageTablePool           = Buffer;
+    mPageTablePool->NextPool = mPageTablePool;
+  } else {
+    ((PAGE_TABLE_POOL *)Buffer)->NextPool = mPageTablePool->NextPool;
+    mPageTablePool->NextPool              = Buffer;
```
os-d: I think this was my comment last time and I apologize, I don't recall the answer, but does it make sense to use the standard list functions here instead of manipulating it in the same fashion?

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  // If we are running low on free pages, allocate more while we
+  // still have enough pages to accommodate a potential split. Hold
+  // a lock during this because the AllocatePages() call in GetMorePages()
+  // may call this function recursively. In that recursive call, it should
+  // pull from the existing pool rather than allocating more pages.
+  if (  ((TotalFreePages < Pages) ||
+         ((TotalFreePages - Pages) <= MIN_PAGES_AVAILABLE))
+     && !mPageTablePoolLock)
```
os-d: nit: some weird spacing happened here

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: If we're going to go below the minimum number of pages reserved by serving the current request, the logic will hold a semaphore and call to allocate more page table memory. That request will require at most 4 more pages if every level needs a table entry and may cause reentry into AllocatePageTableMemory(). Because the semaphore is held, that request will be serviced by the pages still in reserve to prevent infinite recursion. Once the top level request for more translation table memory resumes execution, it will relinquish the lock and have at least another 512 pages for translation table entries.

os-d (approved):

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: Removed

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: Fixed

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: Ah I forgot to add that part.

It took a bit of a rework, but I updated it to use the list functions from BaseLib. I also fixed a very unlikely edge case where multiple pool reserves collectively meet the high-water mark but no one pool can service the page request.

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,202 @@
...
+    DEBUG ((DEBUG_ERROR, "ERROR: Out of aligned pages\n"));
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  // Reserve one page for pool header.
+  Buffer->FreePages = PoolPages - 1;
+  Buffer->Offset    = EFI_PAGE_SIZE;
+  Buffer->Signature = PAGE_TABLE_POOL_SIGNATURE;
+  TotalFreePages   += Buffer->FreePages;
+  InsertHeadList (&mPageTablePoolList, &Buffer->NextPool);
```
os-d, line 70: Do we want to add to the tail of the list? Maybe it doesn't matter, at this point we should have done all operations on the pages backing this, but just want to check if these entries get used immediately if that is ok or not

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,202 @@
...
+  IN UINTN  Pages
+  )
+{
+  PAGE_TABLE_POOL  *PoolToAllocateFrom = NULL;
+  LIST_ENTRY       *CurrentEntry       = mPageTablePoolList.ForwardLink;
+
+  while (CurrentEntry != &mPageTablePoolList) {
+    PoolToAllocateFrom = CR (CurrentEntry, PAGE_TABLE_POOL, NextPool, PAGE_TABLE_POOL_SIGNATURE);
+    if (Pages <= PoolToAllocateFrom->FreePages) {
+      break;
```
os-d, line 94: The high water mark is supposed to prevent this, but do we need an edge case condition if we don't find any pool that has enough pages?

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: If more pages are being reserved then either it's a really large request in which case it would be best to put it at the tail or it's an average sized request and the reserve is almost empty in which case it's better to put it at the head. In the code base currently, every single call to AllocatePageTableMemory() will be requesting one page so it's better to put it at the head to minimize the length of each list walk.

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: The edge cases are handled in AllocatePageTableMemory() when this function returns NULL

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -0,0 +1,202 @@
...
+{
+  PAGE_TABLE_POOL  *Buffer;
+
+  if (PoolPages == 0) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  PoolPages++;
+  PoolPages = ALIGN_VALUE (PoolPages, EFI_SIZE_TO_PAGES (SIZE_2MB)); // Add one page for the header
+  Buffer    = (PAGE_TABLE_POOL *)AllocateAlignedPages (PoolPages, BASE_2MB);
```
kuqin12: nit: why is this BASE_2MB instead of SIZE_2MB?

kuqin12 (approved):

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
TaylorBeebe: I must have searched Base.h for a 2MB size macro two separate times and found the two results. Fixed

PR #257, opened 2024-07-24
Changed all line endings to CRLF

kenlautner:
## Description

Changed all line endings for every file to CRLF.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


os-d (approved):
Does this need to do something to turn LineEndingCheckPlugin on? Or why was this not failing otherwise?

We should add this commit to .git-ignore-rev (or whatever that file is called). @makubacki, can we only do that after this PR goes in, since we don't know the hash of the commit in 202405 until it gets checked in?

.gitmodules
makubacki, line 1: Git files should stay LF.

.markdownlintignore
```diff
@@ -1,9 +1,9 @@
...
-/Common/MU_TIANO
+# Ignore build directory
+/Build/
+
+# Ignore external dependencies
+*_extdep/
+
+# Ignore cloned dependencies
+/Common/MU_TIANO
 /MU_BASECORE
```
makubacki: Since you're already making a line ending only change, can you please add an empty line to the end of any files in the PR missing one?

makubacki (approved):

makubacki:
> We should add this commit to .git-ignore-rev (or whatever that file is called). @makubacki, can we only do that after this PR goes in, since we don't know the hash of the commit in 202405 until it gets checked in?

You should be able to add another commit to the PR branch that updates the ignore rev file with the SHA of the first commit (line ending fixes) and that SHA would be preserved with Rebase & FF. Somewhat fragile since a rebase of the PR branch would change that so a second PR would be fine.

.markdownlintignore
kenlautner: Updated.

.gitmodules
kenlautner: Updated.

PR #265, opened 2024-08-01
[202405] Add Cumulative CodeQL changes.

apop5:
## Description
Cumulative CodeQL fixes for the release 202405 Branch.

9553d450a0

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran local CI under WSL.

## Integration Instructions
New static build assert will occur if a mismatch MM_CALLER_MINOR_VER is integrated at some point in the future.
The fix for this will be to either remove the static assert at that time, or to update the static assert.
This is to alert the builder that a version change has occurred and the code needs examined to determine any impacts.

### DIFF
--- a/.pytool/CISettings.py
+++ b/.pytool/CISettings.py
@@ -146,10 +146,6 @@ class Settings(CiSetupSettingsManager, CiBuildSettingsManager, UpdateSettingsMan
                 scopes += codeql_helpers.get_scopes(self.codeql)
 
                 if self.codeql:
-                    shell_environment.GetBuildVars().SetValue(
-                        "STUART_CODEQL_AUDIT_ONLY",
-                        "TRUE",
-                        "Set in CISettings.py")
                     codeql_filter_files = [str(n) for n in glob.glob(
                         os.path.join(self.GetWorkspaceRoot(),
                                      '**/CodeQlFilters.yml'),
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -259,8 +259,17 @@ GetMmCompatibility (
 
   MmVersion = MmVersionArgs.Arg0;
 
+  // MU_CHANGE - CodeQL Change - unsigned-comparison-zero
+  // Make MM_CALLER_MINOR_VER more restrictive. Rquire
+  //  exact minor version match rather than >=.
+  // Add static_assert to check if/when verison change
+  // occurs.
+  STATIC_ASSERT (
+    MM_CALLER_MINOR_VER == 0,
+    "MM_CALLER_MINOR_VERSION has changed!"
+    );
   if ((MM_MAJOR_VER (MmVersion) == MM_CALLER_MAJOR_VER) &&
-      (MM_MINOR_VER (MmVersion) >= MM_CALLER_MINOR_VER))
+      (MM_MINOR_VER (MmVersion) == MM_CALLER_MINOR_VER))
   {
     DEBUG ((
       DEBUG_INFO,
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -260,10 +260,13 @@ GetMmCompatibility (
   MmVersion = MmVersionArgs.Arg0;
 
   // MU_CHANGE - CodeQL Change - unsigned-comparison-zero
-  // Make MM_CALLER_MINOR_VER more restrictive. Rquire
+  // Make MM_CALLER_MINOR_VER more restrictive. Require
   //  exact minor version match rather than >=.
-  // Add static_assert to check if/when verison change
-  // occurs.
+
+  // Add static_assert to check if/when version change
+  // occurs. If a version change occurs,
+  // the static assert will need to be updated to
+  // match the new major/minor version.
   STATIC_ASSERT (
     MM_CALLER_MINOR_VER == 0,
     "MM_CALLER_MINOR_VERSION has changed!"
--- a/ArmPkg/Drivers/ArmGic/ArmGicLib.c
+++ b/ArmPkg/Drivers/ArmGic/ArmGicLib.c
@@ -246,7 +246,7 @@ ArmGicSetInterruptPriority (
     MmioAndThenOr32 (
       GicDistributorBase + ARM_GIC_ICDIPR + (4 * RegOffset),
       ~(0xff << RegShift),
-      Priority << RegShift
+      (UINT32)(Priority << RegShift) // MU_CHANGE - CodeQL Change
       );
   } else {
     GicCpuRedistributorBase = GicGetCpuRedistributorBase (
@@ -260,7 +260,7 @@ ArmGicSetInterruptPriority (
     MmioAndThenOr32 (
       IPRIORITY_ADDRESS (GicCpuRedistributorBase, RegOffset),
       ~(0xff << RegShift),
-      Priority << RegShift
+      (UINT32)(Priority << RegShift) // MU_CHANGE - CodeQL Change
       );
   }
 }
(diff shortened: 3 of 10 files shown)


.pytool/CISettings.py
```diff
@@ -147,10 +145,6 @@ def GetActiveScopes(self):
                 scopes += codeql_helpers.get_scopes(self.codeql)
 
                 if self.codeql:
-                    shell_environment.GetBuildVars().SetValue(
-                        "STUART_CODEQL_AUDIT_ONLY",
-                        "TRUE",
-                        "Set in CISettings.py")
```
os-d, line 152: I'm assuming this should be the last commit. Also, is removing the imports intended?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -259,9 +259,9 @@ GetMmCompatibility (
 
   MmVersion = MmVersionArgs.Arg0;
 
-  if ((MM_MAJOR_VER (MmVersion) == MM_CALLER_MAJOR_VER) &&
-      (MM_MINOR_VER (MmVersion) >= MM_CALLER_MINOR_VER))
-  {
+  // MU_CHANGE - CodeQL Change - unsigned-comparison-zero
+  // Dropped checking of minor version
+  if ((MM_MAJOR_VER (MmVersion) == MM_CALLER_MAJOR_VER)) {
```
os-d: Is it really valid to drop checking of the minor version?

ArmPkg/Library/DebugPeCoffExtraActionLib/DebugPeCoffExtraActionLib.c
```diff
@@ -32,7 +32,7 @@ PeCoffLoaderRelocateImageExtraAction (
   IN OUT PE_COFF_LOADER_IMAGE_CONTEXT  *ImageContext
   )
 {
-#ifdef __GNUC__
+ #ifdef __GNUC__
   if (ImageContext->PdbPointer) {
     DEBUG ((
       DEBUG_LOAD | DEBUG_INFO,
```
os-d: The changes here do not look CodeQL related, can we drop them?

os-d (approved):

ArmPkg/Library/DebugPeCoffExtraActionLib/DebugPeCoffExtraActionLib.c
apop5: This one is uncrustify. CI failed uncrustify check because of this.

os-d: This is the only change in this file though, how is uncrustify not failing on the repo currently?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
apop5: Yes, I think it is.

Looking at how the versions could rev in the future, if any major version change occurs, it will fail, as it written today
The minor version is always greater than or equal, which covers all possible current changes today.


The alternative is to make it more restrictive, and force the major/minor to exactly match the caller version.

os-d: Well, this is on ARM, where StMM lives in secure world and can be independently updated. So we could have DXE at a newer minor version and StMM at an older version. @kuqin12, do you have an opinion here?

.pytool/CISettings.py
apop5: The things imported are unused in the file, python auto formatter will remove unreferenced imports.

apop5: They can be added back, if needed, but without the imports being referenced, it doesn't provide any benefit to keep them there.

ArmPkg/Library/DebugPeCoffExtraActionLib/DebugPeCoffExtraActionLib.c
apop5: https://github.com/microsoft/mu_silicon_arm_tiano/blob/release/202311/ArmPkg/ArmPkg.ci.yaml#L284

It is failing in current release/202405, but it is in audit only mode.

I can revert the change.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
apop5: Based on offline conversation, the following was agreed:

`strict match + static assert`

.pytool/CISettings.py
apop5: added them back.

kuqin12 (approved):

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -259,8 +259,17 @@ GetMmCompatibility (
 
   MmVersion = MmVersionArgs.Arg0;
 
+  // MU_CHANGE - CodeQL Change - unsigned-comparison-zero
+  // Make MM_CALLER_MINOR_VER more restrictive. Rquire
```
makubacki:
```suggestion
  // Make MM_CALLER_MINOR_VER more restrictive. Require

```

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -259,8 +259,17 @@ GetMmCompatibility (
...
   MmVersion = MmVersionArgs.Arg0;
 
+  // MU_CHANGE - CodeQL Change - unsigned-comparison-zero
+  // Make MM_CALLER_MINOR_VER more restrictive. Rquire
+  //  exact minor version match rather than >=.
+  // Add static_assert to check if/when verison change
+  // occurs.
+  STATIC_ASSERT (
+    MM_CALLER_MINOR_VER == 0,
+    "MM_CALLER_MINOR_VERSION has changed!"
```
makubacki, line 272: Is this a new requirement that the minor version cannot be zero?

makubacki: If so, it should be marked as a breaking change.

makubacki (approved):

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
apop5: This stems from the CodQl error that is was comparing 0 (MM_CALLER_MINOR_VERSION) against greater than or equal to zero, which was a usless comparison.

The minor version has not changed in 6 years, since it was introduced.

Additionally, the >= comparison meant they were trying to make it future minor version compatible, meaning that a minor version update would not cause a break in the functionality.

After discussing with kuqin12, it was decided to make the check more restrictive (meaning it needs to match the version) and to add a static assert to break on a minor version change instead of just showing an error in the debug log.

I can mark this as breaking, but I assume that if it has not changed in 6 years, the chance of it being updated is minimal.

makubacki: I initially misinterpreted the assert logic when looking through quickly earlier. A minor version typically exists to allow version tracking with breaking compatibility, so this seems a little strange in that regard.

Also, the comparison is only considered useless because of the current value of `MM_CALLER_MINOR_VER` and the fact that it's a static value. The original intention is still valid though the context is not apparent to CodeQL.

Is the expectation in the current implementation that the `STATIC_ASSERT` condition is updated over time? If it stays, I think it needs to explain in more detail what the impact and resolution is.

apop5: I updated the comment around the added STATIC_ASSERT to say that it should be updated over time if/when the caller version ever changes.

PR #295, opened 2024-10-04
ArmPkg: Fix call to BuildCpuHob in CpuPei.c

mebersol:
## Description

Fix CpuPei.c for aarch64. Change function signature for ArmGetPhysicalAddressBits.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated handoff on virtual platform bringup

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/CpuPei/CpuPei.c
+++ b/ArmPkg/Drivers/CpuPei/CpuPei.c
@@ -138,7 +138,7 @@ InitializeCpuPeim (
   ArmEnableBranchPrediction ();
 
   // Publish the CPU memory and io spaces sizes
-  BuildCpuHob ((UINT8)ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize)); // MU_CHANGE Add typecast
+  BuildCpuHob (ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize));
 
 
   // Only MP Core platform need to produce gArmMpCoreInfoPpiGuid
--- a/ArmPkg/Drivers/CpuPei/CpuPei.c
+++ b/ArmPkg/Drivers/CpuPei/CpuPei.c
@@ -138,7 +138,7 @@ InitializeCpuPeim (
   ArmEnableBranchPrediction ();
 
   // Publish the CPU memory and io spaces sizes
-  BuildCpuHob (ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize));
+  BuildCpuHob ((UINT8)ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize)); // MU_CHANGE Add typecast
 
   // Only MP Core platform need to produce gArmMpCoreInfoPpiGuid
   Status = PeiServicesLocatePpi (&gArmMpCoreInfoPpiGuid, 0, NULL, (VOID **)&ArmMpCoreInfoPpi);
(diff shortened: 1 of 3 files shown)


VivianNK (approved):

ArmPkg/Drivers/CpuPei/CpuPei.c
```diff
@@ -138,7 +138,7 @@ InitializeCpuPeim (
   ArmEnableBranchPrediction ();
 
   // Publish the CPU memory and io spaces sizes
-  BuildCpuHob (ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize));
+  BuildCpuHob ((UINT8)ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize));
```
VivianNK:
```suggestion
  BuildCpuHob ((UINT8)ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize)); // MU_CHANGE Add typecast

```

os-d (approved):
Approved with Vivian's suggestion

ArmPkg/Drivers/CpuPei/CpuPei.c
```diff
@@ -138,7 +138,8 @@ InitializeCpuPeim (
   ArmEnableBranchPrediction ();
 
   // Publish the CPU memory and io spaces sizes
-  BuildCpuHob (ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize));
+  BuildCpuHob ((UINT8)ArmGetPhysicalAddressBits (), PcdGet8 (PcdPrePiCpuIoSize)); // MU_CHANGE Add typecast
```
kuqin12, line 141: I am sure this change works.

But would it be more appropriate to update the function prototype to return UINT8 for this function (ArmGetPhysicalAddressBits)? Or do we really see some potential to have a 256+bit CPU in the foreseeable future?

os-d: Yeah, I was thinking about that same thing. I'd support that approach.

VivianNK (approved):

ArmPkg/Include/Library/ArmLib.h
makubacki, line 1: Is there a reason not to take this to edk2 and get the MdePkg owner approval first?

https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Library/ArmLib.h#L680

makubacki, line 1: As discussed offline, should be marked as a breaking change since this changes a library public API.

os-d: I was considering this also, was willing to take it here first because it is blocking a platform integration of 2405 do to MSVC build failure

makubacki: Okay, urgency is usually the reason. I think the approval chance is high. I'll go ahead and approve.

makubacki: Actually, is the original issue in [ArmPkg/Drivers/CpuPei/CpuPei.c](https://github.com/microsoft/mu_silicon_arm_tiano/pull/295/files/795fea0abbc280a6eb8f3c41e64235670725bf62#diff-a83e707f83046888d95a18832c426c7ad6ab95d878f07e3896561e51e23f821b) the only problem?

If so, we can fix the build for now with that change which is a fix, non-breaking, and non-diverging with the public API and then propose the API change with more time and come back to update the CpuPei.c file, right?

os-d: That’s fine with me, too. That is the only problem. That’s how Mike had it, Kun asked for the more general expansion. Let’s go with the simple cast and we’ll take up the breaking change

mebersol: Thanks for the feedback; back to the original.

PR #305, opened 2024-11-08
StandaloneMmPkg: Arm/StandaloneMmCoreEntryPoint Remove Check for SMC ID

os-d:
## Description

Today, StandaloneMmCoreEntryPoint checks for an SMC EventId of MM_COMMUNICATE or FF-A Direct and drops all other SMCs. However, the TCG ACPI spec dictates that a different SMC ID, platform defined, will be send to communicate with the TPM[1]. When a platform is using StandaloneMM and the SPM_MM configuration, this check causes this valid SMC to be dropped. This is an issue because TPM calls cannot be routed, including TPM calls originating from the OS.

This patch drops the check to allow StandaloneMmCoreEntryPoint to route the calls to StandaloneMmCpu to decide whether it can handle the SMC (which on platforms with this TPM configuration it can).

"This field provides the SMC/HVC call function ID that will invoke the TPM start method.
Firmware SHALL implement the SMC call as an SMC32 or SMC64 Fast Call, compliant with the SMC Calling convention
 specification.  The call takes no parameters, no
client ID, no Secure OS ID, and no Session ID. The call SHALL return zero. The function ID SHALL be allocated from a Service Call Range over which the platform vendor has authority."

[1]: https://trustedcomputinggroup.org/wp-content/uploads/TCG_ACPIGeneralSpecification_v1p3_r6_14april2021.pdf Section 8.3.1

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on platform using TPM SMCs.

## Integration Instructions

N/A.

### DIFF
--- a/ArmPkg/Library/StandaloneMmCoreEntryPoint/Arm/StandaloneMmCoreEntryPoint.c
+++ b/ArmPkg/Library/StandaloneMmCoreEntryPoint/Arm/StandaloneMmCoreEntryPoint.c
@@ -141,80 +141,68 @@ DelegatedEventLoop (
     DEBUG ((DEBUG_INFO, "X6 :  0x%x\n", (UINT32)EventCompleteSvcArgs->Arg6));
     DEBUG ((DEBUG_INFO, "X7 :  0x%x\n", (UINT32)EventCompleteSvcArgs->Arg7));
 
-    //
-    // ARM TF passes SMC FID of the MM_COMMUNICATE interface as the Event ID upon
-    // receipt of a synchronous MM request. Use the Event ID to distinguish
-    // between synchronous and asynchronous events.
-    //
-    if ((ARM_SMC_ID_MM_COMMUNICATE != (UINT32)EventCompleteSvcArgs->Arg0) &&
-        (ARM_SVC_ID_FFA_MSG_SEND_DIRECT_REQ != (UINT32)EventCompleteSvcArgs->Arg0))
-    {
-      DEBUG ((DEBUG_ERROR, "UnRecognized Event - 0x%x\n", (UINT32)EventCompleteSvcArgs->Arg0));
-      Status = EFI_INVALID_PARAMETER;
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

ArmPkg/Library/StandaloneMmCoreEntryPoint/Arm/StandaloneMmCoreEntryPoint.c
```diff
@@ -141,80 +141,68 @@ DelegatedEventLoop (
...
-    // ARM TF passes SMC FID of the MM_COMMUNICATE interface as the Event ID upon
-    // receipt of a synchronous MM request. Use the Event ID to distinguish
-    // between synchronous and asynchronous events.
-    //
-    if ((ARM_SMC_ID_MM_COMMUNICATE != (UINT32)EventCompleteSvcArgs->Arg0) &&
-        (ARM_SVC_ID_FFA_MSG_SEND_DIRECT_REQ != (UINT32)EventCompleteSvcArgs->Arg0))
-    {
-      DEBUG ((DEBUG_ERROR, "UnRecognized Event - 0x%x\n", (UINT32)EventCompleteSvcArgs->Arg0));
-      Status = EFI_INVALID_PARAMETER;
+    FfaEnabled = FeaturePcdGet (PcdFfaEnable);
```
apop5, line 144: Should this have mu_change tags?

apop5 (approved):

ArmPkg/Library/StandaloneMmCoreEntryPoint/Arm/StandaloneMmCoreEntryPoint.c
os-d: Fair enough, I just cp'd my edk2 PR. Will update

PR #311, opened 2024-11-19
[Rebase & FF][FF-A] Adding FDT based MemoryAllocationLib and ServicesTableLib for Secure Partitions

kuqin12:
## Description

This change adds a new library instance to support heap allocation management for secure partitions.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on feature/ffa_enablement branch with QEMU SBSA platform.

## Integration Instructions

- The platform needs to link this new library as the memory allocation library for secure partition builds.
- In the secure partition DTS, adding a memory region for heap will allow the memory allocation library to allocate memory from this region.

### DIFF
--- a/ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
+++ b/ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
@@ -20,11 +20,11 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/ArmMmuLib.h>
 #include <Library/ArmSvcLib.h>
 #include <Library/DebugLib.h>
-#include <Library/HobLib.h>
 #include <Library/BaseLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/SerialPortLib.h>
 #include <Library/ArmStandaloneMmMmuLib.h>
+#include <Library/SecurePartitionServicesTableLib.h>
 #include <Library/PcdLib.h>
 
 #include <IndustryStandard/ArmStdSmc.h>
@@ -32,24 +32,14 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <IndustryStandard/ArmFfaSvc.h>
 #include <IndustryStandard/ArmFfaBootInfo.h>
 
-#define SPM_MAJOR_VER_MASK   0xFFFF0000
-#define SPM_MINOR_VER_MASK   0x0000FFFF
-#define SPM_MAJOR_VER_SHIFT  16
-
-#define SPM_MAJOR_VER  0
-#define SPM_MINOR_VER  1
-
-#define BOOT_PAYLOAD_VERSION  1
-
 #define FFA_PAGE_4K   0
 #define FFA_PAGE_16K  1
 #define FFA_PAGE_64K  2
 
-// Local variable to help Standalone MM Core decide whether FF-A ABIs can be
-// used for all communication. This variable is usable only after the StMM image
-// has been relocated and all image section permissions have been correctly
-// updated.
-STATIC BOOLEAN  mUseOnlyFfaAbis = FALSE;
+// Materialize the Secure Partition Services Table
+SECURE_PARTITION_SERVICES_TABLE  mSpst = {
+  .FDTAddress = NULL
+};
 
 /**
   This structure is used to stage boot information required to initialize the
@@ -336,7 +326,6 @@ ModuleEntryPoint (
   INT32                         Ret;
   UINT32                        SectionHeaderOffset;
   UINT16                        NumberOfSections;
-  VOID                          *HobStart;
   VOID                          *TeData;
   UINTN                         TeDataSize;
   EFI_PHYSICAL_ADDRESS          ImageBase;
@@ -411,6 +400,10 @@ ModuleEntryPoint (
     goto finish;
   }
 
+  // Now that we can update globals, initialize the SPST for other libraries
+  mSpst.FDTAddress = DtbAddress;
+  gSpst            = &mSpst;
+
   if (ImageContext.ImageAddress != (UINTN)TeData) {
     ImageContext.ImageAddress = (UINTN)TeData;
     ArmSetMemoryRegionNoExec (ImageBase, SIZE_4KB);
@@ -420,15 +413,14 @@ ModuleEntryPoint (
     ASSERT_EFI_ERROR (Status);
   }
 
-  // Update the global copy now that the image has been relocated.
-  mUseOnlyFfaAbis = UseOnlyFfaAbis;
+  ProcessLibraryConstructorList (NULL, NULL);
 
   ProcessLibraryConstructorList (NULL, NULL);
 
   //
   // Call the MM Core entry point
   //
-  ProcessModuleEntryPointList (HobStart);
+  ProcessModuleEntryPointList (NULL);
 
 finish:
   if (Status == RETURN_UNSUPPORTED) {
--- a/ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
+++ b/ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
@@ -1,0 +1,399 @@
+/** @file
+  MM Memory page management functions.
+
+  Copyright (c) 2009 - 2014, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2016 - 2018, ARM Limited. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiMm.h>
+
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+
+typedef struct {
+  LIST_ENTRY    Link;
+  UINTN         NumberOfPages;
+} FREE_PAGE_LIST;
+
+#define NEXT_MEMORY_DESCRIPTOR(MemoryDescriptor, Size) \
+  ((EFI_MEMORY_DESCRIPTOR *)((UINT8 *)(MemoryDescriptor) + (Size)))
+
+#define TRUNCATE_TO_PAGES(a)  ((a) >> EFI_PAGE_SHIFT)
+
+LIST_ENTRY  mMmMemoryMap = INITIALIZE_LIST_HEAD_VARIABLE (mMmMemoryMap);
+
+UINTN  mMapKey;
+
+/**
+  Internal Function. Allocate n pages from given free page node.
+
+  @param  Pages                  The free page node.
+  @param  NumberOfPages          Number of pages to be allocated.
+  @param  MaxAddress             Request to allocate memory below this address.
+
+  @return Memory address of allocated pages.
+
+**/
+UINTN
+InternalAllocPagesOnOneNode (
+  IN OUT FREE_PAGE_LIST  *Pages,
+  IN     UINTN           NumberOfPages,
+  IN     UINTN           MaxAddress
+  )
+{
+  UINTN           Top;
+  UINTN           Bottom;
+  FREE_PAGE_LIST  *Node;
+
+  Top = TRUNCATE_TO_PAGES (MaxAddress + 1 - (UINTN)Pages);
+  if (Top > Pages->NumberOfPages) {
+    Top = Pages->NumberOfPages;
+  }
+
+  Bottom = Top - NumberOfPages;
+
+  if (Top < Pages->NumberOfPages) {
+    Node                = (FREE_PAGE_LIST *)((UINTN)Pages + EFI_PAGES_TO_SIZE (Top));
+    Node->NumberOfPages = Pages->NumberOfPages - Top;
+    InsertHeadList (&Pages->Link, &Node->Link);
+  }
+
+  if (Bottom > 0) {
+    Pages->NumberOfPages = Bottom;
+  } else {
+    RemoveEntryList (&Pages->Link);
+  }
+
+  return (UINTN)Pages + EFI_PAGES_TO_SIZE (Bottom);
+}
+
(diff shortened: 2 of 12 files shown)


ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
```diff
@@ -0,0 +1,399 @@
...
+    if ((UINTN)Pages <= Address) {
+      if ((UINTN)Pages + EFI_PAGES_TO_SIZE (Pages->NumberOfPages) < EndAddress) {
+        break;
+      }
+
+      return InternalAllocPagesOnOneNode (Pages, NumberOfPages, EndAddress);
+    }
+  }
+
+  return ~Address;
```
Raymond-MS, line 141: This seems like an odd return value, what was the decision behind returning the inverted address and not -1 like in other functions?

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
```diff
@@ -0,0 +1,399 @@
...
+MmInternalAllocatePages (
+  IN  EFI_ALLOCATE_TYPE     Type,
+  IN  EFI_MEMORY_TYPE       MemoryType,
+  IN  UINTN                 NumberOfPages,
+  OUT EFI_PHYSICAL_ADDRESS  *Memory
+  )
+{
+  UINTN  RequestedAddress;
+
+  if ((MemoryType != EfiRuntimeServicesCode) &&
```
Raymond-MS, line 171: This parameter is only checked here but not used anywhere else in the function, why are we passing it in?

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Pool.c
```diff
@@ -0,0 +1,329 @@
...
+
+  ASSERT ((FreePoolHdr->Header.Size & (FreePoolHdr->Header.Size - 1)) == 0);
+  ASSERT (((UINTN)FreePoolHdr & (FreePoolHdr->Header.Size - 1)) == 0);
+  ASSERT (FreePoolHdr->Header.Size >= MIN_POOL_SIZE);
+
+  PoolIndex                     = (UINTN)(HighBitSet32 ((UINT32)FreePoolHdr->Header.Size) - MIN_POOL_SHIFT);
+  FreePoolHdr->Header.Available = TRUE;
+  ASSERT (PoolIndex < MAX_POOL_INDEX);
+  InsertHeadList (&mMmPoolLists[PoolIndex], &FreePoolHdr->Link);
+  return EFI_SUCCESS;
```
Raymond-MS, line 186: Any reason for this function to return a status if it will always return SUCCESS? We should just be able to make it a void return.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
```diff
@@ -0,0 +1,1051 @@
...
+VOID
+EFIAPI
+FreePages (
+  IN VOID   *Buffer,
+  IN UINTN  Pages
+  )
+{
+  EFI_STATUS  Status;
+
+  ASSERT (Pages != 0);
```
Raymond-MS, line 149: I see we do an assert on pages to verify it is a non-zero value, why don't we also check if Buffer is non-null? That or we could at least just return immediately if so.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
```diff
@@ -0,0 +1,1051 @@
...
+VOID
+EFIAPI
+FreeAlignedPages (
+  IN VOID   *Buffer,
+  IN UINTN  Pages
+  )
+{
+  EFI_STATUS  Status;
+
+  ASSERT (Pages != 0);
```
Raymond-MS, line 356: Same question here about checking Buffer. :)

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
```diff
@@ -0,0 +1,1051 @@
...
+  IN EFI_MEMORY_TYPE  PoolType,
+  IN UINTN            OldSize,
+  IN UINTN            NewSize,
+  IN VOID             *OldBuffer  OPTIONAL
+  )
+{
+  VOID  *NewBuffer;
+
+  NewBuffer = InternalAllocateZeroPool (PoolType, NewSize);
+  if ((NewBuffer != NULL) && (OldBuffer != NULL)) {
```
Raymond-MS, line 704: This is just a question really, what happens if OldBuffer is not passed in (since it is marked OPTIONAL), does it get defaulted to NULL?

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
```diff
@@ -0,0 +1,1051 @@
...
+  @param  MemoryType            The type of memory to allocate.
+  @param  Pages                 The number of 4 KB pages to allocate.
+  @param  Alignment             The requested alignment of the allocation.  Must be a power of two.
+                                If Alignment is zero, then byte alignment is used.
+
+  @return A pointer to the allocated buffer or NULL if allocation fails.
+
+**/
+VOID *
+InternalAllocateAlignedPages (
```
Raymond-MS, line 174: Are "Internal" functions not supposed to be marked STATIC or are they not really internal? Just a question. :)

Raymond-MS:
In general, since this PR has a large change in it, I would recommend another set of eyes. :)

Raymond-MS (approved):

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
```diff
@@ -423,12 +422,12 @@ ModuleEntryPoint (
   // Update the global copy now that the image has been relocated.
   mUseOnlyFfaAbis = UseOnlyFfaAbis;
 
-  ProcessLibraryConstructorList (NULL, NULL);
+  ProcessLibraryConstructorList (DtbAddress, NULL);
```
apop5: I have concerns about how ImageHandle is being taken to pass DtbAddress.

Taking ImageHandle and using it in this manor seems like we are going to cause more problems down the line when we encounter a library constructor that attempts to use the passed values as they are stated in the prototypes.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
kuqin12: This is probably a quick change to the input value. The important characteristic is that upon such failure, the return has to be different than the input value. A constant value will not work. We can do +1, -1 or other for the same effect.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Page.c
kuqin12: This is just saying that we will only allow runtime type.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Pool.c
kuqin12: Actually a nice call. We probably should convert these asserts into error codes instead.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
kuqin12: There is no strict definition of "Internal" here. But in this case, it means it might be used outside of this file but should stay in the module.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
kuqin12: Hmm, can we really rule out the Buffer never being null? It is unorthodox. But the actual usage will depend on what regions this SP will have access to?

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
kuqin12: Oh, optional means it could be NULL when caller invokes this function. Not that caller does not need to pass in anything for this argument. If I understand you correctly?

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
kuqin12: Yeah. I was thinking about the same. The hard part is that this FDT address will be automatically populated by the loader in the input parameters at the entrypoint function. We could save it to a global variable and have others access it that way. But since I imagine this will be used in other libraries as well, using an extern will be a bad pattern. Any suggestion is welcome.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
Raymond-MS: It is entirely possible that buffer will never be NULL but it never hurts to put in safe code to make sure we aren't doing something we don't want. We check Pages so I don't see why we shouldn't also check Buffer. It should always be the case that if Pages !=0 then Buffer != NULL.

apop5 (approved):

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
apop5: summarizing our offline conversation:

UefiBootServicesTableLib and UefiRuntimeServicesTableLib create gBS and gRT global variables, and their implementations are provided by the library instance.
Would doing something similar for DtbAddress be acceptable?
This entry point can still access it directly, but to allow library constructors to access it, they would consume the library instance directly.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
kuqin12: Updated.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/SecurePartitionMemoryAllocationLib.c
kuqin12: Updated.

ArmPkg/Library/SecurePartitionMemoryAllocationLib/Pool.c
kuqin12: Fixed.

ArmPkg/Library/SecurePartitionServicesTableLib/SecurePartitionServicesTableLib.inf
```diff
@@ -0,0 +1,39 @@
+## @file
+# Module entry point library for DXE core.
+#
+# Copyright (c) 2017 - 2021, Arm Ltd. All rights reserved.<BR>
```
apop5:
```suggestion
# Copyright (c) Microsoft Corporation.

```

ArmPkg/Library/SecurePartitionServicesTableLib/SecurePartitionServicesTableLib.inf
```diff
@@ -0,0 +1,39 @@
+## @file
+# Module entry point library for DXE core.
```
apop5:
```suggestion
# Secure Partition System Table library implementation.

```

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
```diff
@@ -311,7 +302,7 @@ GetSpManifest (
   DEBUG ((DEBUG_ERROR, "SP manifest not found \n"));
   return EFI_NOT_FOUND;
 }
-
+extern volatile BOOLEAN loop;
```
apop5: left over?

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
apop5: Changes look good to me.

ArmPkg/Library/SecurePartitionEntryPoint/StandaloneMmCoreEntryPoint.c
kuqin12: Yep, removed.

ArmPkg/Library/SecurePartitionServicesTableLib/SecurePartitionServicesTableLib.inf
kuqin12: Thanks, updated.

ArmPkg/Library/SecurePartitionServicesTableLib/SecurePartitionServicesTableLib.inf
kuqin12: Updated.

PR #321, opened 2024-12-17
[REBASE][FF-A] Ported MS Services to UEFI Style

Raymond-MS:
## Description

All services from Trusted Services have been ported to UEFI style and are running in a secure partition generated our UEFI style as well. The services have been ported as libraries that secure partitions can include if they would like to utilize them. Added an ArmArchTimer library for the secure partition to support the TPM library for the TPM service. SEL0 SP's do not have access to timer registers so we needed a work-around.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified service functionality through the FfaPartitionTestApp with the TPM service being tested through building with the TPM defines set in the build_conf file. Secure partition is being dispatched and all services are being exercised. 

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Include/Library/NotificationServiceLib.h
+++ b/ArmPkg/Include/Library/NotificationServiceLib.h
@@ -1,0 +1,96 @@
+/** @file
+  Definitions for the Notification Service
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef NOTIFICATION_SERVICE_LIB_H_
+#define NOTIFICATION_SERVICE_LIB_H_
+
+#include <Base.h>
+#include <IndustryStandard/ArmFfaPartInfo.h>
+#include <Library/ArmSvcLib.h>
+#include <Library/ArmFfaLibEx.h>
+
+typedef INT32 NotificationStatus;
+
+#define NOTIFICATION_STATUS_SUCCESS            ((NotificationStatus)0)
+#define NOTIFICATION_STATUS_GENERIC_ERROR      ((NotificationStatus)-1)
+#define NOTIFICATION_STATUS_INVALID_PARAMETER  ((NotificationStatus)-2)
+
+#define NOTIFICATION_OPCODE_BASE     (0)
+#define NOTIFICATION_OPCODE_QUERY    (NOTIFICATION_OPCODE_BASE + 0)
+#define NOTIFICATION_OPCODE_SETUP    (NOTIFICATION_OPCODE_BASE + 1)
+#define NOTIFICATION_OPCODE_DESTROY  (NOTIFICATION_OPCODE_BASE + 2)
+
+#define NOTIFICATION_SERVICE_UUID \
+{ \
+    0xb510b3a3, 0x59f6, 0x4054, { 0xba, 0x7a, 0xff, 0x2e, 0xb1, 0xea, 0xc7, 0x65 } \
+}
+
+/**
+  Initializes the Notification service
+
+**/
+VOID
+NotificationServiceInit (
+  VOID
+  );
+
+/**
+  Deinitializes the Notification service
+
+**/
+VOID
+NotificationServiceDeInit (
+  VOID
+  );
+
+/**
+  Handler for Notification service commands
+
+  @param  Request   The incoming message
+  @param  Response  The outgoing message
+
+**/
+VOID
+NotificationServiceHandle (
+  DIRECT_MSG_ARGS_EX  *Request,
+  DIRECT_MSG_ARGS_EX  *Response
+  );
+
+/**
+  Calls NotificationSet on the given ID with the given flag
+
+  @param  Id           The ID to trigger the event on
+  @param  ServiceUuid  The service containing the ID to trigger
+  @param  Flag         The NotificationSet flag to use
+
+  @retval NOTIFICATION_STATUS_SUCCESS           Success
+  @retval NOTIFICATION_STATUS_INVALID_PARAMETER Invalid parameter
+  @retval NOTIFICATION_STATUS_GENERIC_ERROR     NotificationSet failed
+
+**/
+NotificationStatus
+NotificationServiceIdSet (
+  UINT16  Id,
+  UINT8   *ServiceUuid,
+  UINT32  Flag
+  );
+
+/**
+  Extracts the UUID from the message arguments
+
+  @param  Request   The incoming message to extract the UUID from
+  @param  Uuid      The UUID to populate
+
+**/
+VOID
+NotificationServiceExtractUuid (
+  DIRECT_MSG_ARGS_EX  *Request,
+  UINT8               *Uuid
+  );
+
+#endif /* NOTIFICATION_SERVICE_LIB_H_ */
--- a/ArmPkg/Include/Library/TpmServiceLib.h
+++ b/ArmPkg/Include/Library/TpmServiceLib.h
@@ -1,0 +1,48 @@
+/** @file
+  Definitions for the TPM Service
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef TPM_SERVICE_LIB_H_
+#define TPM_SERVICE_LIB_H_
+
+#include <Base.h>
+#include <IndustryStandard/ArmFfaPartInfo.h>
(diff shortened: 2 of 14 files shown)


ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,333 @@
...
+VOID
+TpmServiceDeInit (
+  VOID
+  )
+{
+  /* Nothing to DeInit */
+}
+
+VOID
+TpmServiceHandle (
```
kuqin12, line 435: I think these libraries are well architected to be in separate commits.

Let's break this PR into 3 commits (maybe 4 if we stick to the timer lib you have here), one for each service.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+STATIC
+TpmStatus
+FinishHandler (
+  DIRECT_MSG_ARGS_EX *Request, 
+  DIRECT_MSG_ARGS_EX *Response
+  )
+{
+  (void)Request;
```
kuqin12: What is this line for?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+STATIC
+TpmStatus
+UnregisterHandler (
+  DIRECT_MSG_ARGS_EX *Request, 
+  DIRECT_MSG_ARGS_EX *Response
+  )
+{
+  (void)Request;
```
kuqin12: Same here?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+STATIC
+TpmStatus
+RegisterHandler (
+  DIRECT_MSG_ARGS_EX *Request, 
+  DIRECT_MSG_ARGS_EX *Response
+  )
+{
+  (void)Request;
```
kuqin12: ditto

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
kuqin12: Also do you need these handler related functions in this context?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,333 @@
...
+/* TPM Service Defines */
+#define TPM_MAJOR_VER              (1)
+#define TPM_MINOR_VER              (0)
+
+#define TPM_START_PROCESS_CMD      (0)
+#define TPM_START_PROCESS_LOC_REQ  (1)
+
+#define TPM_LOC_STATE_MASK         (0x9F)
+#define TPM_LOC_STS_MASK           (0x01)
+#define TPM_CTRL_START_MASK        (0x01)
```
kuqin12: nit: do you think it makes sense to have these live in the lib header?

Raymond-MS: No because they aren't needed outside of the .c file. These are only needed by the cleanup function which makes sure the CRB is in a good state.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Removed. This was due to an unused parameter in TS and was copied here.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Removed

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Removed. And yes, I think it is worth keeping them stubbed out in case we need to implement them in the future. This doesn't hurt and keeps things clear rather than just returning an error code in the handler.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -0,0 +1,471 @@
...
+  )
+{
+  /* TODO: If we end up supporting more than Locality0, this will need to init all localities supported */
+  PTP_CRB_REGISTERS_PTR  InternalTpmCrb;
+
+  InternalTpmCrb = (PTP_CRB_REGISTERS_PTR)(UINTN)PcdGet64 (PcdTpmInternalBaseAddress);
+  DEBUG ((DEBUG_INFO, "PcdTpmInternalBaseAddress: %lx\n", PcdGet64 (PcdTpmInternalBaseAddress)));
+  SetMem ((void *)InternalTpmCrb, sizeof (PTP_CRB_REGISTERS), 0x00);
+  InternalTpmCrb->LocalityState = 0x01;   // Default Value - tpmEstablished
+  InternalTpmCrb->InterfaceId   = 0x4011; // CRB Interface Selected, Only Locality0 Supported
```
kuqin12: Consider using #define?

ArmPkg/Include/Library/TpmServiceLib.h
```diff
@@ -0,0 +1,77 @@
...
+#define TPM_STATUS_INV_CRB_CTRL_DATA  ((TpmStatus)0x8e000006)
+#define TPM_STATUS_ALREADY            ((TpmStatus)0x8e000009)
+#define TPM_STATUS_DENIED             ((TpmStatus)0x8e00000A)
+#define TPM_STATUS_NOMEM              ((TpmStatus)0x8e00000B)
+
+/* Opcodes
+ * TPM Opcodes spec defined in: https://developer.arm.com/documentation/den0138/0100/?lang=en */
+#define TPM_OPCODE_BASE                   (0x0F000000)
+#define TPM_OPCODE_GET_INTERFACE_VERSION  (TPM_OPCODE_BASE + 0x01)
+#define TPM_OPCODE_GET_FEATURE_VERSION    (TPM_OPCODE_BASE + 0x101)
```
kuqin12: This is already defined in the `Guid/Tpm2ServiceFfa.h`. Probably should reuse those?

ArmPkg/Include/Library/NotificationServiceLib.h
```diff
@@ -0,0 +1,96 @@
...
+
+**/
+
+#ifndef NOTIFICATION_SERVICE_LIB_H_
+#define NOTIFICATION_SERVICE_LIB_H_
+
+#include <Base.h>
+#include <IndustryStandard/ArmFfaPartInfo.h>
+#include <Library/ArmSvcLib.h>
+#include <Library/ArmFfaLibEx.h>
```
kuqin12, line 15: nit: we generally do not recommend carrying header files in the header files. But we can worry about it later.

ArmPkg/Library/NotificationServiceLib/NotificationServiceLib.c
```diff
@@ -0,0 +1,518 @@
...
+} NotifBits;
+
+typedef struct  {
+  UINT8        ServiceUuid[16];
+  NotifBits    ServiceBits[NOTIFICATION_MAX_IDS];
+  BOOLEAN      InUse;
+} NotifService;
+
+/* Notification Service Variables */
+static BOOLEAN       IdsAcquired                                     = FALSE;
```
kuqin12: Use `STATIC`?

ArmPkg/Library/ArmArchTimerLibEx/ArmArchTimerLibEx.inf
```diff
@@ -0,0 +1,32 @@
+#/** @file
+#
+#  Copyright (c) 2011 - 2014, ARM Limited. All rights reserved.<BR>
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#**/
+
```
kuqin12, line 7: Probably want to add an override tag for this:https://github.com/microsoft/mu_basecore/blob/dev/202405/BaseTools/Plugin/OverrideValidation/ReadMe.md

kuqin12 (approved):

PR #329, opened 2025-01-21
[FF-A][TPM] Updated TPM Service for Proper State Transitions

Raymond-MS:
## Description

Implemented a translation library for the TPM service to translate the Internal PC CRB to an external TPM type (CRB/FIFO). Updated the TPM service to support going through each phase of the TPM (cmdReady/Start/goIdle). Streamlined the Internal CRB clean-up. Updated the FF-A TPM library to go through the normal state procedure (cmdReady/Start/goIdle).

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified by running the build with TPM2_ENABLE=TRUE. Verified with the FfaPartitionTestApp.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
+++ b/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
@@ -1,0 +1,93 @@
+/** @file
+  Definitions for the TPM Service State Translation Library
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef TPM_SST_LIB_H_
+#define TPM_SST_LIB_H_
+
+#include <IndustryStandard/TpmPtp.h>
+
+/**
+  Initiates the transition to the Idle state
+
+  @param  Locality The locality of the TPM to set into Idle
+
+  @retval EFI_SUCCESS  Success
+  @retval EFI_TIMEOUT  Timeout
+
+**/
+EFI_STATUS
+TpmSstGoIdle (
+  UINT8 Locality
+  );
+
+/**
+  Initiates the transition to the commandReady state
+
+  @param  Locality  The locality of the TPM to set to commandReady
+
+  @retval EFI_SUCCESS  Success
+  @retval EFI_TIMEOUT  Timeout
+
+**/
+EFI_STATUS
+TpmSstCmdReady (
+  UINT8 Locality
+  );
+
+/**
+  Initiates command execution
+
+  @param  Locality        The locality of the TPM to initiate the command on
+  @param  InternalTpmCrb  The internal CRB to copy command data from
+
+  @retval EFI_SUCCESS  Success
+  @retval EFI_TIMEOUT  Timeout
+
+**/
+EFI_STATUS
+TpmSstStart (
+  UINT8                 Locality,
+  PTP_CRB_REGISTERS_PTR InternalTpmCrb
+  );
+
+/**
+  Requests access to the given locality
+
+  @param  Locality  The locality to request access to
+
+  @retval EFI_SUCCESS  Success
+  @retval EFI_TIMEOUT  Timeout
+
+**/
+EFI_STATUS
+TpmSstLocalityRequest (
+  UINT8 Locality
+  );
+
+/**
+  Returns if IdleBypass is supported
+
+  @retval TRUE   Supported
+  @retval FALSE  Unsupported
+
+**/
+BOOLEAN
+TpmSstIsIdleBypassSupported (
+  VOID
+  );
+
+/**
+  Initializes the TPM Service State Translation Library
+
+**/
+VOID
+TpmSstInit(
+  VOID
+  );
+
+#endif /* TPM_SST_LIB_H_ */
--- a/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
+++ b/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
@@ -1,5 +1,9 @@
 /** @file
-  Definitions for the TPM Service State Translation Library
+  Definitions for the TPM Service State Translation Library. This library's
+  main purpose is to translate the states of the TPM service's CRB states to
+  the main TPM's interface states. (i.e. TPM PC CRB -> TPM FIFO) A user of the
+  TPM service should only need to update this library with the proper TPM
+  interface type for their device.
 
   Copyright (c), Microsoft Corporation.
   SPDX-License-Identifier: BSD-2-Clause-Patent
--- a/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
+++ b/ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
@@ -22,7 +22,7 @@
 **/
 EFI_STATUS
 TpmSstGoIdle (
-  UINT8 Locality
+  UINT8  Locality
   );
 
 /**
@@ -36,7 +36,7 @@ TpmSstGoIdle (
 **/
 EFI_STATUS
 TpmSstCmdReady (
-  UINT8 Locality
+  UINT8  Locality
   );
 
 /**
@@ -51,8 +51,8 @@ TpmSstCmdReady (
 **/
 EFI_STATUS
 TpmSstStart (
-  UINT8                 Locality,
-  PTP_CRB_REGISTERS_PTR InternalTpmCrb
+  UINT8                  Locality,
+  PTP_CRB_REGISTERS_PTR  InternalTpmCrb
   );
 
 /**
@@ -66,7 +66,7 @@ TpmSstStart (
 **/
 EFI_STATUS
 TpmSstLocalityRequest (
-  UINT8 Locality
+  UINT8  Locality
   );
 
 /**
@@ -86,7 +86,7 @@ TpmSstIsIdleBypassSupported (
 
 **/
 VOID
-TpmSstInit(
+TpmSstInit (
   VOID
   );
 
--- a/ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
+++ b/ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
@@ -1,5 +1,16 @@
 /** @file
-  Implementation for the TPM Service
+  Implementation for the TPM Service. This library is based
+  off of the ARM spec: TPM Service Command Response Buffer
+  Interface Over FF-A. The spec can be found here:
+
+  https://developer.arm.com/documentation/den0138/0100/?lang=en
+
+  The state flow is based off the TCG PC Client Specific Platform
+  TPM Profile for TPM 2.0. The spec can be found here:
+
+  https://trustedcomputinggroup.org/resource/pc-client-platform-tpm-profile-ptp-specification/
+
+  Figure 4 - TPM State Diagram for CRB Interface
 
   Copyright (c), Microsoft Corporation.
   SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -13,6 +24,7 @@
 #include <Library/TpmServiceStateTranslationLib.h>
 #include <Guid/Tpm2ServiceFfa.h>
 #include <IndustryStandard/TpmPtp.h>
+#include <IndustryStandard/Tpm20.h>
 
 /* TPM Service Defines */
 #define TPM_MAJOR_VER  (1)
@@ -21,14 +33,7 @@
 #define TPM_START_PROCESS_CMD      (0)
 #define TPM_START_PROCESS_LOC_REQ  (1)
 
-#define TPM_CTRL_REQ_CMD_READY_MASK  (0x01)
-#define TPM_CTRL_REQ_GO_IDLE_MASK    (0x02)
-#define TPM_CTRL_STS_IDLE_MASK       (0x02)
-
 #define TPM_LOCALITY_OFFSET  (0x1000)
-
-#define TPM_LOC_STATE_DEFAULT     (0x01)   // Default Value - tpmEstablished
-#define TPM_INTERFACE_ID_DEFAULT  (0x4111) // Default Value - CRB Interface, all localities supported
 
 /* TPM Service States */
 typedef enum {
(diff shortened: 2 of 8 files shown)


ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.inf
```diff
@@ -0,0 +1,42 @@
...
+  SecurityPkg/SecurityPkg.dec
+
+[LibraryClasses]
+  BaseLib
+  BaseMemoryLib
+  IoLib
+  TimerLib
+  DebugLib
+  PlatformFfaInterruptLib
+  ArmSvcLib
```
kuqin12: Do you need all these libraries here?

ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
```diff
@@ -0,0 +1,93 @@
+/** @file
+  Definitions for the TPM Service State Translation Library
```
kuqin12: Just curious, why is it "state translation"? It is more than "translation", but also committing the message, right?

kuqin12: I think it makes sense to have some more description on the rationale of this library creation. I was going to call it TpmHalLib, but that might be too vague.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
+  NUM_TPM_STATES
+} TpmState;
+
+/* TPM Service Localities */
+typedef enum {
+  TPM_LOCALITY_0 = 0,
+  TPM_LOCALITY_1,
+  TPM_LOCALITY_2,
+  TPM_LOCALITY_3,
+  TPM_LOCALITY_4,
```
kuqin12: Hmm, I think this is making the assumption that you are supporting all 5 localities. Do we reference the 5 pages here implicitly?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
+  NUM_TPM_LOCALITIES
```
kuqin12: btw, i just noticed that `NUM_LOCALITIES` is defined in the `Tpm20.h`

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
-// Default Value - CRB Interface Selected, Only Locality0 Supported
-#define TPM_INTERFACE_ID_DEFAULT  (0x4011)
+#define TPM_CTRL_REQ_CMD_READY_MASK  (0x01)
+#define TPM_CTRL_REQ_GO_IDLE_MASK    (0x02)
+#define TPM_CTRL_STS_IDLE_MASK       (0x02)
+
+#define TPM_LOCALITY_OFFSET  (0x1000)
+
+#define TPM_LOC_STATE_DEFAULT     (0x01)   // Default Value - tpmEstablished
+#define TPM_INTERFACE_ID_DEFAULT  (0x4111) // Default Value - CRB Interface, all localities supported
```
kuqin12: Hmm, I think it is might better to use a `PTP_CRB_INTERFACE_IDENTIFIER` struct and populate the sub fields? 0x4111 probably encoded too much implicit information :)

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
-#define TPM_LOC_STS_MASK     (0x01)
-#define TPM_CTRL_START_MASK  (0x01)
-
-// Default Value - tpmEstablished
-#define TPM_LOC_STATE_DEFAULT  (0x01)
-// Default Value - CRB Interface Selected, Only Locality0 Supported
-#define TPM_INTERFACE_ID_DEFAULT  (0x4011)
+#define TPM_CTRL_REQ_CMD_READY_MASK  (0x01)
+#define TPM_CTRL_REQ_GO_IDLE_MASK    (0x02)
+#define TPM_CTRL_STS_IDLE_MASK       (0x02)
```
kuqin12: Are these

PTP_CRB_CONTROL_AREA_REQUEST_COMMAND_READY
PTP_CRB_CONTROL_AREA_REQUEST_GO_IDLE
PTP_CRB_CONTROL_AREA_STATUS_TPM_IDLE

respectively?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -75,72 +97,93 @@ ConvertEfiToTpmStatus (
...
-  VOID
-  )
-{
-  /* TODO: If we end up supporting more than Locality0, this will need to clean the correct locality registers */
-  PTP_CRB_REGISTERS_PTR  InternalTpmCrb;
+  /* Set the locality state based on the active locality. */
+  InternalTpmCrb->LocalityState = TPM_LOC_STATE_DEFAULT;
+  switch (ActiveLocality) {
+    case TPM_LOCALITY_0:
+      InternalTpmCrb->LocalityState |= PTP_CRB_LOCALITY_STATE_ACTIVE_LOCALITY_0;
```
kuqin12, line 140: Why is this "bit or"?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
+  TPM_LOCALITY_2,
+  TPM_LOCALITY_3,
+  TPM_LOCALITY_4,
+  NUM_TPM_LOCALITIES
+} TpmLocality;
 
 typedef UINTN TpmStatus;
 
+/* TPM Service Variables */
+static TpmState  CurrentState;
```
kuqin12:
```suggestion
STATIC TpmState  mCurrentState;
```

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -21,17 +21,39 @@
...
+static UINT8     ActiveLocality;
```
kuqin12:
```suggestion
STATIC UINT8     mActiveLocality;
```

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -158,53 +201,120 @@ HandleCommand (
...
+      if (InternalTpmCrb->CrbControlRequest & TPM_CTRL_REQ_GO_IDLE_MASK) {
+        DEBUG ((DEBUG_INFO, "READY State - Handle TPM Command goIdle Request\n"));
+        Status = TpmSstGoIdle (ActiveLocality);
+        if (Status == EFI_SUCCESS) {
+          CurrentState = TPM_STATE_IDLE;
+        }
+
+        /* Check the cmdReady bit in the CrbControlRequest register, clear it if it has been
+         * set again. */
+      } else if (InternalTpmCrb->CrbControlRequest & TPM_CTRL_REQ_CMD_READY_MASK) {
```
kuqin12: For this case, you do not transit state?

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.inf
```diff
@@ -0,0 +1,42 @@
...
+#
+#  Copyright (c), Microsoft Corporation.
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#**/
+
+[Defines]
+  INF_VERSION                    = 1.29
+  BASE_NAME                      = TpmServiceStateTranslationLib
```
kuqin12, line 13: I think this is the Tis specific instance? If so, you might want to add that to the name. There might be a ptp instance for the same library class.

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.c
```diff
@@ -0,0 +1,646 @@
...
+
+#define LOCALITY_OFFSET  (0x1000)
+
+#define DELAY_AMOUNT  (30)
+
+#define PTP_TIMEOUT_MAX  (90000 * 1000) // 90s
+
+/* TPM Service State Translation Library Variables */
+STATIC BOOLEAN  IsCrbInterface;
+STATIC BOOLEAN  IsIdleBypassSupported;
```
kuqin12: We normally have a "m" as the prefix of globals.

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.c
```diff
@@ -0,0 +1,646 @@
+/** @file
+  Implementation for the TPM Service State Translation Library
+
+  Copyright (c), Microsoft Corporation.
```
kuqin12, line 10: If there is code inherited from open source version, please include the original copyright.

kuqin12: The same for other files involved in this PR.

kuqin12: The same for other files involved in this PR.

kuqin12:
I think the change overall makes sense. Two requests would be:
1. Can we add a state machine flowchart? That will help to inspect the edge cases.
2. Add some host-based test for the handlecommand function. (this does not need to come in the same PR)

ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
Raymond-MS: The library is mainly there to translate the different states the TPM goes through (IDLE, READY, COMPLETE). This is why it exposes GoIdle, CommandReady, and Start. The only thing not really related to states is RequestLocality. I didn't want to call it HAL because it isn't really a hardware abstraction layer. Did you want me to remove "State"? I was thinking of doing so but figured it wouldn't be worth the effort. Up to you, feel free to make the call. :)

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Yes, will remove and use those instead. Thanks!

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Sure, I can do that. :)

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: As far as I know, the plan is to support all localities which is why I made this change. I believe so, yes. The localities are sequential.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Doing just a "=" would overwrite the default above.

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: No, we are already in the READY state.

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.inf
Raymond-MS: This is not TIS specific. The library supports both TIS and CRB translation.

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.inf
Raymond-MS: Probably not, removed.

ArmPkg/Library/TpmServiceStateTranslationLib/TpmServiceStateTranslationLib.inf
kuqin12: Hmm, do you prefer to carry the extra weight for this service? I guess this library covers the Fifo and Crb case. But that may/will not be all. i.e. it might talk to another fTPM.

Probably want to add some specifier for this instance anyway.

kuqin12 (approved):

ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
```diff
@@ -0,0 +1,97 @@
...
+/**
+  Initiates the transition to the Idle state
+
+  @param  Locality The locality of the TPM to set into Idle
+
+  @retval EFI_SUCCESS  Success
+  @retval EFI_TIMEOUT  Timeout
+
+**/
+EFI_STATUS
```
apop5, line 27: With this being a library interface, should these functions have an EFIAPI calling convention?

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
```diff
@@ -158,53 +197,120 @@ HandleCommand (
...
-                         &ResponseDataLen,
-                         TpmCommandBuffer
-                         );
+      break;
 
-  /* Copy the response data from the static buffer */
-  CopyMem (InternalTpmCrb->CrbDataBuffer, TpmCommandBuffer, ResponseDataLen);
+    /* The proper state flow should be: IDLE -> READY -> START -> IDLE/READY. */
+    default:
+      break;
```
apop5, line 305: If the state gets out of sorts, should default be used as a way to catch it? Or recover?
i.e. should a debug message be added and/or mCurrentState get reset to something

apop5 (approved):

ArmPkg/Library/TpmServiceLib/TpmServiceLib.c
Raymond-MS: Yes, you are right. It shouldn't be the case but if for some reason the state gets corrupted, I should handle it gracefully. Good catch. Thanks!

ArmPkg/Include/Library/TpmServiceStateTranslationLib.h
Raymond-MS: As per our discussion, going to leave this as is for now.

PR #344, opened 2025-02-24
[Rebase & FF] Add 18 register support for SMC and SVC calls for AArch64 platforms

kuqin12:
## Description

Current SMC and SVC only supports up to 8 registers.

However, the FF-A spec has definition for FFA direct request 2 which could use up to 18 registers.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on proprietary hardware and QEMU SBSA virtual platform.

## Integration Instructions

For platforms needing to use 18-registers, add `gArmTokenSpaceGuid.PcdSxcUse18Registers|TRUE` to the platform build dsc file.

### DIFF
--- a/ArmPkg/Include/Library/ArmFfaLib.h
+++ b/ArmPkg/Include/Library/ArmFfaLib.h
@@ -42,8 +42,37 @@ typedef struct DirectMsgArgs {
   /// Implementation define argument 3, this will be set to/from x6(v1) or x7(v2)
   UINTN    Arg3;
 
-  /// Implementation define argument 4, this will be set to/from x7(v1) or ignored (v2)
+  /// Implementation define argument 4, this will be set to/from x7(v1) or x8(v2)
   UINTN    Arg4;
+
+  // MU_CHANGE Starts: Add implementation define arguments 5-13
+  /// Implementation define argument 5, this will be set to/from x9(v2)
+  UINTN    Arg5;
+
+  /// Implementation define argument 6, this will be set to/from x10(v2)
+  UINTN    Arg6;
+
+  /// Implementation define argument 7, this will be set to/from x11(v2)
+  UINTN    Arg7;
+
+  /// Implementation define argument 8, this will be set to/from x12(v2)
+  UINTN    Arg8;
+
+  /// Implementation define argument 9, this will be set to/from x13(v2)
+  UINTN    Arg9;
+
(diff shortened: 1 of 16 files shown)


kuqin12:
Upstream link is: https://github.com/tianocore/edk2/pull/10685

Raymond-MS (approved):
This looks good to me. I think the changes need MU_CHANGE tags though?

kuqin12:
> This looks good to me. I think the changes need MU_CHANGE tags though?

Alright. Added those tags... Well, as good as I can. Some assembly files are tricky.

ArmPkg/Include/Library/ArmFfaLib.h
```diff
@@ -42,8 +42,37 @@ typedef struct DirectMsgArgs {
   /// Implementation define argument 3, this will be set to/from x6(v1) or x7(v2)
   UINTN    Arg3;
 
-  /// Implementation define argument 4, this will be set to/from x7(v1) or ignored (v2)
+  /// Implementation define argument 4, this will be set to/from x7(v1) or x8(v2)
   UINTN    Arg4;
+
+  // MU_CHANGE Starts: Add implementation define arguments 5-13
+  /// Implementation define argument 5, this will be set to/from x9(v2)
+  UINTN    Arg5;
```
eeshanl, line 50: Here and below, would it be better to have these args as an array instead? that way in the future more can be added with ease. 
Not sure how easy that would be to integrate with existing code that is already doing ->ArgN for example. Would have to get all users of the ArmFfaLib to use the array Arg[N] instead.

ArmPkg/Include/Library/ArmSmcLib.h
```diff
@@ -23,6 +23,18 @@ typedef struct {
   UINTN    Arg5;
   UINTN    Arg6;
   UINTN    Arg7;
+  // MU_CHANGE Starts: Support 18 arguments for AArch64
+  UINTN    Arg8;
```
eeshanl, line 27: same as above

ArmPkg/Include/Library/ArmFfaLib.h
eeshanl: per teams chat, keeping as is. So LGTM.

eeshanl (approved):

PR #354, opened 2025-03-19
[SQUASH ON REBASE] SmmuDxe updates and bug fixes

eeshanl:
### [SQUASH ON REBASE] SmmuDxe updates and bug fixes

Bug fix for command queue address not being aligned properly.
Bug fix for event queue consumption.
Dynamically sets starting level of translation for paging from IDR5.OAS on the smmu.
Improve logging on error.
Safety improvements on updating page table entries in the IoMmu protocol.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU SBSA platform as well as a physical ARM surface platform

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -56,14 +56,12 @@ UpdateFlags (
   IN UINT32      Index
   )
 {
-  EFI_STATUS  Status;
-  UINT64      Entry;
+  UINT64  Entry;
 
   if ((Table == NULL) || ((Flags & ~PAGE_TABLE_BLOCK_OFFSET) != 0) || (Index >= PAGE_TABLE_SIZE)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter.\n", __func__));
-    Status = EFI_INVALID_PARAMETER;
-    ASSERT_EFI_ERROR (Status);
-    return Status;
+    ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
+    return EFI_INVALID_PARAMETER;
   }
 
   // Allows clearing the R/W bits without affecting the other bits in the entry.
@@ -266,7 +264,6 @@ IoMmuMap (
   EFI_STATUS            Status;
   EFI_PHYSICAL_ADDRESS  PhysicalAddress;
   IOMMU_MAP_INFO        *MapInfo;
-  SMMUV3_CMD_GENERIC    Command;
 
   if ((This == NULL) ||
       (HostAddress == NULL) ||
@@ -288,30 +285,11 @@ IoMmuMap (
   }
 
   // Invalidate TLBI Command
-  SMMUV3_BUILD_CMD_TLBI_NSNH_ALL (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
+  Status = SmmuV3TLBInvalidateAll (mSmmu);
   if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_TLBI_NSNH_ALL failed.\n", __func__));
+    DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
     goto Error;
   }
-
-  SMMUV3_BUILD_CMD_TLBI_EL2_ALL (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_TLBI_EL2_ALL failed.\n", __func__));
-    goto Error;
-  }
-
-  // Issue a CMD_SYNC command to guarantee that any previously issued TLB
-  // invalidations (CMD_TLBI_*) are completed (SMMUv3.2 spec section 4.6.3).
-  SMMUV3_BUILD_CMD_SYNC_NO_INTERRUPT (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_SYNC_NO_INTERRUPT failed.\n", __func__));
-    goto Error;
-  }
-
-  ArmDataSynchronizationBarrier ();
 
   // Allocate and fill the IOMMU_MAP_INFO structure with mapped information
   *DeviceAddress = PhysicalAddress; // Identity mapping
@@ -350,9 +328,8 @@ IoMmuUnmap (
   IN  VOID                  *Mapping
   )
 {
-  EFI_STATUS          Status;
-  SMMUV3_CMD_GENERIC  Command;
-  IOMMU_MAP_INFO      *MapInfo;
+  EFI_STATUS      Status;
+  IOMMU_MAP_INFO  *MapInfo;
 
   if ((This == NULL) || (Mapping == NULL)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
@@ -369,30 +346,11 @@ IoMmuUnmap (
   }
 
   // Invalidate TLBI Command
-  SMMUV3_BUILD_CMD_TLBI_NSNH_ALL (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
+  Status = SmmuV3TLBInvalidateAll (mSmmu);
   if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_TLBI_NSNH_ALL failed.\n", __func__));
+    DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
     goto Error;
   }
-
-  SMMUV3_BUILD_CMD_TLBI_EL2_ALL (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_TLBI_EL2_ALL failed.\n", __func__));
-    goto Error;
-  }
-
-  // Issue a CMD_SYNC command to guarantee that any previously issued TLB
-  // invalidations (CMD_TLBI_*) are completed (SMMUv3.2 spec section 4.6.3).
-  SMMUV3_BUILD_CMD_SYNC_NO_INTERRUPT (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: CMD_SYNC_NO_INTERRUPT failed.\n", __func__));
-    goto Error;
-  }
-
-  ArmDataSynchronizationBarrier ();
 
   // Free the mapping structure allocated in IoMmuMap
   if (MapInfo != NULL) {
@@ -530,9 +488,8 @@ IoMmuSetAttribute (
   IN UINT64                IoMmuAccess
   )
 {
-  EFI_STATUS          Status;
-  IOMMU_MAP_INFO      *MapInfo;
-  SMMUV3_CMD_GENERIC  Command;
+  EFI_STATUS      Status;
+  IOMMU_MAP_INFO  *MapInfo;
 
   if ((This == NULL) || (Mapping == NULL) || ((IoMmuAccess & ~(EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE)) != 0)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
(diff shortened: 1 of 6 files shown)


os-d:
Mark as [SQUASH ON REBASE]? I'm assuming that for upstreaming to edk2 and carrying forward in Mu branches we would just want this squashed with the original SMMU driver commit?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -65,24 +66,18 @@ UpdateFlags (
     return Status;
   }
 
-  Status = EFI_SUCCESS;
```
Raymond-MS, line 68: Small Nit: Since this was removed and you are returning EFI_SUCCESS directly, it looks as though you can remove the variable entirely and just use EFI_INVALID_PARAMETER above. :)

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -65,24 +66,18 @@ UpdateFlags (
...
-    if (Flags != 0) {
-      // Set R/W bits in page table entry
-      Table->Entries[Index] |= Flags;
-    } else {
-      // Clear R/W bits in page table entry
-      Table->Entries[Index] &= ~(PAGE_TABLE_READ_BIT | PAGE_TABLE_WRITE_BIT);
-    }
-  } else {
+  if (Flags != 0) {
     // Set R/W bits in page table entry
```
os-d: It could be setting other bits, too, right?

Raymond-MS (approved):
LGTM

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Done

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -233,7 +235,10 @@ SmmuV3AllocateEventQueue (
 
   *QueueLog2Size = MIN (Idr1.EventQs, SMMUV3_EVENT_QUEUE_LOG2ENTRIES);
   QueueSize      = SMMUV3_EVENT_QUEUE_SIZE_FROM_LOG2 (*QueueLog2Size);
-  return AllocateZeroPool (QueueSize);
+  Pages          = EFI_SIZE_TO_PAGES (QueueSize);
+  EventQueueBase = AllocatePages (Pages);
+  ZeroMem (EventQueueBase, EFI_PAGES_TO_SIZE (Pages));
```
spbrogan: what if Allocation fails?

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -263,7 +270,10 @@ SmmuV3AllocateCommandQueue (
 
   *QueueLog2Size = MIN (Idr1.CmdQs, SMMUV3_COMMAND_QUEUE_LOG2ENTRIES);
   QueueSize      = SMMUV3_COMMAND_QUEUE_SIZE_FROM_LOG2 (*QueueLog2Size);
-  return AllocateZeroPool (QueueSize);
+  Pages          = EFI_SIZE_TO_PAGES (QueueSize);
+  CmdQueueBase   = AllocatePages (Pages);
```
spbrogan: same concern here.  Need error handling

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: Good catch, fixed. I check if the cmd/event queues are null after these functions are called, but I also ZeroMem it here without error checking.

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: done same as above

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -211,59 +211,87 @@ PageTableDeInit (
...
+    return EFI_INVALID_PARAMETER;
   }
 
   Idr1.AsUINT32 = SmmuV3ReadRegister32 (SmmuInfo->SmmuBase, SMMU_IDR1);
 
   *QueueLog2Size = MIN (Idr1.CmdQs, SMMUV3_COMMAND_QUEUE_LOG2ENTRIES);
   QueueSize      = SMMUV3_COMMAND_QUEUE_SIZE_FROM_LOG2 (*QueueLog2Size);
-  return AllocateZeroPool (QueueSize);
+  Pages          = EFI_SIZE_TO_PAGES (QueueSize);
+  *CmdQueueBase  = AllocatePages (Pages);
```
kuqin12, line 286: Did I remember right that these needs to be aligned? There is also allocatealignedpages API, I think.

kuqin12 (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -142,19 +136,10 @@ UpdateMapping (
       }
 
       ZeroMem ((VOID *)NewPage, EFI_PAGE_SIZE);
+      WriteBackInvalidateDataCacheRange (NewPage, EFI_PAGE_SIZE);
```
os-d: Are we not cache coherent at this point?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -170,23 +155,26 @@ UpdateMapping (
...
-        Current->Entries[Index]  = (PhysicalAddress & ~PAGE_TABLE_BLOCK_OFFSET); // Assign PA
-        Current->Entries[Index] |= PAGE_TABLE_ENTRY_VALID_BIT;                   // valid entry
+        Entry = (PhysicalAddress & ~PAGE_TABLE_BLOCK_OFFSET); // Assign PA
+        // validate entry and set leaf level flags
+        Entry                  |= PAGE_TABLE_ACCESS_FLAG | PAGE_TABLE_DESCRIPTOR | PAGE_TABLE_ENTRY_VALID_BIT;
+        Current->Entries[Index] =  Entry;
       } else {
-        Current->Entries[Index] &= ~PAGE_TABLE_ENTRY_VALID_BIT; // only invalidate leaf entry
+        Entry                   = Current->Entries[Index] & ~PAGE_TABLE_ENTRY_VALID_BIT; // invalidate entry
+        Current->Entries[Index] = Entry;                                                 // only invalidate leaf entry
```
os-d: Looks like this is the same?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -290,21 +277,23 @@ IoMmuMap (
...
-  Status          = UpdatePageTable (mSmmu->PageTableRoot, PhysicalAddress, *NumberOfBytes, Flags, TRUE, FALSE);
+  Status          = UpdatePageTable (mSmmu->PageTableRoot, PhysicalAddress, *NumberOfBytes, 0, TRUE, FALSE);
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a: Failed to update page table.\n", __func__));
     goto Error;
   }
 
-  *DeviceAddress = PhysicalAddress; // Identity mapping
+  // Invalidate TLBI Command
+  Status = SmmuV3TLBInvalidateAll (mSmmu);
```
os-d: Does the whole TLB need to be invalidated? We should be okay with only invalidating this VA?

Also, should the invalidate happen in UpdatePageTable instead of here? Unless that page table code is shared with the CPU page table.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -358,26 +346,9 @@ IoMmuUnmap (
   }
 
   // Invalidate TLBI Command
-  SMMUV3_BUILD_CMD_TLBI_NSNH_ALL (&Command);
-  Status = SmmuV3SendCommand (mSmmu, &Command);
+  Status = SmmuV3TLBInvalidateAll (mSmmu);
```
os-d: Same question, can we just invalidate the VA?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -541,6 +512,13 @@ IoMmuSetAttribute (
     goto Error;
   }
 
+  // Invalidate TLBI Command
+  Status = SmmuV3TLBInvalidateAll (mSmmu);
```
os-d: Same

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -108,6 +108,42 @@ SmmuV3EncodeAddressWidth (
...
+  )
+{
+  if ((OutputAddressWidth > 48) || (OutputAddressWidth < 32)) {
+    DEBUG ((DEBUG_ERROR, "%a: OutputAddressWidth not supported.\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Per the Arm ARM VMSA spec, >= 44 bits of address width requires 4 level paging.
+  // Otherwise, 3 level paging is used.
+  if (OutputAddressWidth >= 44) {
```
os-d: Use a define for `44`?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -51,38 +52,30 @@ STATIC
...
-      Table->Entries[Index] |= Flags;
-    } else {
-      // Clear R/W bits in page table entry
-      Table->Entries[Index] &= ~(PAGE_TABLE_READ_BIT | PAGE_TABLE_WRITE_BIT);
-    }
-  } else {
+  if (Flags != 0) {
     // Set R/W bits in page table entry
-    Table->Entries[Index] |= Flags;
+    Entry = Table->Entries[Index] | Flags;
```
cfernald: Should this not ensure that they are cleared first. i.e. if the caller provides 0b10 but its currently set to 0b01 is the expectation that is should then be 0b11 or 0b10?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: Yeah, did we ever establish if this was actually needed?

ArmPkg/Drivers/SmmuDxe/SmmuDxe.inf
```diff
@@ -36,6 +36,7 @@
   ArmLib
   BaseLib
   BaseMemoryLib
+  CacheMaintenanceLib
```
cfernald: I wonder if we can drop this since it _should_ be cache coherent, right?

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -108,6 +108,42 @@ SmmuV3EncodeAddressWidth (
...
+**/
+EFI_STATUS
+SmmuV3SetTranslationStartingLevel (
+  IN SMMU_INFO  *SmmuInfo,
+  IN UINT32     OutputAddressWidth,
+  OUT UINT64    *S2Sl0
+  )
+{
+  if ((OutputAddressWidth > 48) || (OutputAddressWidth < 32)) {
+    DEBUG ((DEBUG_ERROR, "%a: OutputAddressWidth not supported.\n", __func__));
```
cfernald: nit: should probably print the address width with this.

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: the requirement is to be aligned to 4096, so AllocatePages should suffice.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: The aim of this code path is to support setting the R/W bits from within IoMmuSetAttributes() . Updating function name to be more clear.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: removing, as is not needed.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: reverting back to what it was before for this line.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: In my testing, I tried to use CMD_TLBI_S2_IPA commands to invalidate the region that I am mapping, however that didn't seem to work on the physical platform. Will put a TODO to look into this some more.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: same as above

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: same as above

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: added defines

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Fixed.

PR #360, opened 2025-04-03
[SQUASH ON REBASE] Add support for N SMMU's and dynamically parse any IORT config

eeshanl:
## Description

Supports N number of SMMU's
Dynamically parses any generic IORT structure and configures the SMMU's accordingly

- [x] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an ARM surface platform and QEMU SBSA.

See https://github.com/microsoft/mu_tiano_platforms/pull/1085 for platform side changes.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -120,7 +120,7 @@ UpdateMapping (
   Current = Root;
 
   // Traverse the page table to the leaf level
-  for (Level = mSmmu->TranslationStartingLevel; Level < PAGE_TABLE_DEPTH - 1; Level++) {
+  for (Level = mIoMmu->SmmuInfo->TranslationStartingLevel; Level < PAGE_TABLE_DEPTH - 1; Level++) {
     Index = PAGE_TABLE_INDEX (VirtualAddress, Level);
 
     if (Current->Entries[Index] == 0) {
@@ -202,6 +202,7 @@ UpdatePageTable (
   EFI_STATUS            Status;
   EFI_PHYSICAL_ADDRESS  PhysicalAddressEnd;
   EFI_PHYSICAL_ADDRESS  CurPhysicalAddress;
+  UINT32                SmmuIndex;
 
   if ((Root == NULL) || ((Flags & ~PAGE_TABLE_BLOCK_OFFSET) != 0) || (PhysicalAddress == 0) || (Bytes == 0)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
@@ -223,10 +224,12 @@ UpdatePageTable (
   }
 
   // Invalidate TLBI Command
-  Status = SmmuV3TLBInvalidateAll (mSmmu);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
-    goto Error;
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    Status = SmmuV3TLBInvalidateAll (&mIoMmu->SmmuInfo[SmmuIndex]);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
+      goto Error;
+    }
   }
 
   return Status;
@@ -265,6 +268,7 @@ IoMmuMap (
   EFI_STATUS            Status;
   EFI_PHYSICAL_ADDRESS  PhysicalAddress;
   IOMMU_MAP_INFO        *MapInfo;
+  UINT32                SmmuIndex;
 
   if ((This == NULL) ||
       (HostAddress == NULL) ||
@@ -279,7 +283,7 @@ IoMmuMap (
   }
 
   PhysicalAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
-  Status          = UpdatePageTable (mSmmu->PageTableRoot, PhysicalAddress, *NumberOfBytes, 0, TRUE, FALSE);
+  Status          = UpdatePageTable (mIoMmu->SmmuInfo->PageTableRoot, PhysicalAddress, *NumberOfBytes, 0, TRUE, FALSE);
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a: Failed to update page table.\n", __func__));
     goto Error;
@@ -295,11 +299,17 @@ IoMmuMap (
   *Mapping                 = MapInfo;
 
   // Only prints errors if Event Queue is not empty and GError != 0
-  SmmuV3LogErrors (mSmmu);
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    SmmuV3LogErrors (&mIoMmu->SmmuInfo[SmmuIndex]);
+  }
+
   return Status;
 
 Error:
-  SmmuV3LogErrors (mSmmu);
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    SmmuV3LogErrors (&mIoMmu->SmmuInfo[SmmuIndex]);
+  }
+
   ASSERT_EFI_ERROR (Status);
   return Status;
 }
@@ -324,6 +334,7 @@ IoMmuUnmap (
 {
   EFI_STATUS      Status;
   IOMMU_MAP_INFO  *MapInfo;
+  UINT32          SmmuIndex;
 
   if ((This == NULL) || (Mapping == NULL)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
@@ -333,7 +344,7 @@ IoMmuUnmap (
 
   MapInfo = (IOMMU_MAP_INFO *)Mapping;
 
-  Status = UpdatePageTable (mSmmu->PageTableRoot, MapInfo->PhysicalAddress, MapInfo->NumberOfBytes, 0, FALSE, FALSE);
+  Status = UpdatePageTable (mIoMmu->SmmuInfo->PageTableRoot, MapInfo->PhysicalAddress, MapInfo->NumberOfBytes, 0, FALSE, FALSE);
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a: Failed to update page table.\n", __func__));
     goto Error;
@@ -345,11 +356,17 @@ IoMmuUnmap (
   }
 
   // Only prints errors if Event Queue is not empty and GError != 0
-  SmmuV3LogErrors (mSmmu);
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    SmmuV3LogErrors (&mIoMmu->SmmuInfo[SmmuIndex]);
+  }
+
   return Status;
 
 Error:
-  SmmuV3LogErrors (mSmmu);
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    SmmuV3LogErrors (&mIoMmu->SmmuInfo[SmmuIndex]);
+  }
+
   ASSERT_EFI_ERROR (Status);
   return Status;
 }
@@ -477,6 +494,7 @@ IoMmuSetAttribute (
 {
   EFI_STATUS      Status;
   IOMMU_MAP_INFO  *MapInfo;
+  UINT32          SmmuIndex;
 
   if ((This == NULL) || (Mapping == NULL) || ((IoMmuAccess & ~(EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE)) != 0)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
@@ -487,7 +505,7 @@ IoMmuSetAttribute (
   MapInfo = (IOMMU_MAP_INFO *)Mapping;
 
   Status = UpdatePageTable (
-             mSmmu->PageTableRoot,
+             mIoMmu->SmmuInfo->PageTableRoot,
              MapInfo->PhysicalAddress,
              MapInfo->NumberOfBytes,
              PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS (IoMmuAccess),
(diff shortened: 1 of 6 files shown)


ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -223,10 +224,12 @@ UpdatePageTable (
   }
 
   // Invalidate TLBI Command
-  Status = SmmuV3TLBInvalidateAll (mSmmu);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
-    goto Error;
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
```
cfernald: Should we optimize this to only invalidate when we are breaking existing mappings? i.e. when `valid == FALSE`

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,443 @@ SmmuV3TLBInvalidateAll (
...
+  if (SmmuInfoArray == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to allocate memory for SMMU info array\n", __func__));
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  // Allocate memory for SMMU node pointers (for output reference lookup)
+  SmmuNodePtrs = AllocateZeroPool (SmmuNodeCount * sizeof (VOID *));
+  if (SmmuNodePtrs == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to allocate memory for SMMU node pointers\n", __func__));
+    FreePool (SmmuInfoArray);
```
cfernald: A general way to handle this without having to worry about freeing the things that have been specifically allocated so far would be to at the beginning of the routine do
```
SmmuInfoArray = NULL;
SmmuNodePtrs  = NULL;
```
Then in the teardown do
```
if (SmmuInfoArray != NULL) {
  FreePool (SmmuInfoArray);
}
if (SmmuNodePtrs != NULL) {
  FreePool (SmmuNodePtrs);
}
```
Then you can just use `goto Error` for all errors. Less error prone IMO.

cfernald (dismissed):

cfernald:
For breaking changes, might be good to update the integration instructions to mention the change to the HOB structure.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

eeshanl:
> For breaking changes, might be good to update the integration instructions to mention the change to the HOB structure.

done, added to readme.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,459 @@ SmmuV3TLBInvalidateAll (
...
+
+  DEBUG ((DEBUG_VERBOSE, "%a: Found %d SMMU nodes\n", __func__, *SmmuNodeCount));
+
+  return EFI_SUCCESS;
+}
+
+/*
+* Get the max stream ID for each SMMU.
+*
+* @param [in]  IortTable      Pointer to the IORT table.
```
cfernald, line 980: missing params and return

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,459 @@ SmmuV3TLBInvalidateAll (
...
+          for (SmmuIndex = 0; SmmuIndex < SmmuNodeCount; SmmuIndex++) {
+            if (OutputNode == SmmuNodePtrs[SmmuIndex]) {
+              // This ID mapping references an SMMU node
+              // Calculate the Stream ID range
+              StartId = IdMapping[IdMappingIndex].OutputBase;
+              EndId   = StartId + IdMapping[IdMappingIndex].NumIds;
+
+              // Store the Stream ID range information
+              for (CurID = StartId; CurID <= EndId; CurID++) {
+                // SmmuInfoArray[k].StreamEntryConfig[CurID].CacheCoherentAttribute = StreamEntryConfig.CacheCoherentAttribute;
```
cfernald: Can you remove these comments? Or fix up their variable usage and add a comment on why the are present but commented

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -295,11 +302,17 @@ IoMmuMap (
   *Mapping                 = MapInfo;
 
   // Only prints errors if Event Queue is not empty and GError != 0
-  SmmuV3LogErrors (mSmmu);
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
```
Raymond-MS, line 303: There should be a way to avoid some of these duplicate lines of code. In certain cases they are doing similar things but with an assert. Per our discussion, this should be updated in various places, if possible. :)

eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -1056,27 +1038,46 @@ InitializeSmmuDxe (
...
+    DEBUG ((DEBUG_ERROR, "%a: Failed to initialize IoMmu Config\n", __func__));
+    return Status;
   }
 
+  IortData = (VOID *)((UINT8 *)SmmuConfig + SmmuConfig->IortOffset);
+
+  Status = SmmuV3ParseIort (IortData, &mIoMmu->SmmuInfo, &mIoMmu->SmmuCount);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to parse IORT for SMMU\n", __func__));
+    return EFI_NOT_FOUND;
```
Raymond-MS: Curious, what is the reason this has to return an explicit EFI_NOT_FOUND rather than the return Status?

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: removed, my mistake.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -956,31 +931,35 @@ SmmuV3ExitBootServices (
...
 
-  Status = SmmuV3DisableTranslation (mSmmu->SmmuBase);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: Failed to disable smmu translation.\n", __func__));
-    ASSERT_EFI_ERROR (Status);
-  }
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    Status = SmmuV3DisableTranslation (mIoMmu->SmmuInfo[SmmuIndex].SmmuBase);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to disable smmu translation.\n", __func__));
```
Raymond-MS, line 954: Curious, what happens if one of these do fail to disable translation or set global bypass? Would it be left in a weird state even though we continue through the other SMMUs?

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -897,48 +863,57 @@ SmmuInit (
...
   }
 
-  Status = SmmuV3DisableTranslation (SmmuInfo->SmmuBase);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: Failed to disable SMMUv3 translation\n", __func__));
-  }
+  for (SmmuIndex = 0; SmmuIndex < IoMmu->SmmuCount; SmmuIndex++) {
+    Status = SmmuV3DisableTranslation (IoMmu->SmmuInfo[SmmuIndex].SmmuBase);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to disable SMMUv3 translation 0x%llx\n", __func__, IoMmu->SmmuInfo[SmmuIndex].SmmuBase));
```
Raymond-MS, line 881: Should there be asserts added here if the deinitialization of any of the SMMUs fail?

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,467 @@ SmmuV3TLBInvalidateAll (
...
+}
+
+/**
+ * Get the number of SMMUV3 nodes in the IORT table.
+ *
+ * @param [in]  IortTable      Pointer to the IORT table.
+ * @param [out] SmmuNodeCount  Pointer to store the number of SMMU nodes found.
+ *
+ * @retval EFI_SUCCESS            Success.
+ * @retval EFI_INVALID_PARAMETER  Invalid Parameters.
```
Raymond-MS, line 932: Missing EFI_NOT_FOUND as a retval.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,467 @@ SmmuV3TLBInvalidateAll (
...
+    DEBUG ((DEBUG_ERROR, "%a: Invalid parameters\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Iort = (EFI_ACPI_6_0_IO_REMAPPING_TABLE *)IortTable;
+  Node = (EFI_ACPI_6_0_IO_REMAPPING_NODE *)((UINT8 *)Iort + Iort->NodeOffset);
+
+  for (Count = 0; Count < Iort->NumNodes; Count++) {
+    if ((Node->Type == EFI_ACPI_IORT_TYPE_ROOT_COMPLEX) || (Node->Type == EFI_ACPI_IORT_TYPE_NAMED_COMP)) {
+      if (Node->NumIdMappings > 0) {
```
Raymond-MS: I think you can avoid an indention here by moving this if up to the previous one:

i.e.:
if (((Node->Type == EFI_ACPI_IORT_TYPE_ROOT_COMPLEX) || (Node->Type == EFI_ACPI_IORT_TYPE_NAMED_COMP)) && (Node->NumIdMappings > 0))

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,467 @@ SmmuV3TLBInvalidateAll (
...
+/**
+ * Collect Stream ID specific information for each SMMU.
+ *
+ * @param [in]  IortTable      Pointer to the IORT table.
+ * @param [in]  SmmuNodePtrs   Pointer to the array of SMMU node pointers.
+ * @param [in]  SmmuNodeCount  Number of SMMU nodes.
+ * @param [out] SmmuInfoArray  Pointer to the array of SMMU_INFO structures.
+ *
+ * @retval EFI_SUCCESS            Success.
+ * @retval EFI_INVALID_PARAMETER  Invalid Parameters.
```
Raymond-MS, line 1083: Missing EFI_NOT_FOUND as a retval.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -869,3 +870,467 @@ SmmuV3TLBInvalidateAll (
...
+    DEBUG ((DEBUG_ERROR, "%a: Invalid parameters\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Iort = (EFI_ACPI_6_0_IO_REMAPPING_TABLE *)IortTable;
+  Node = (EFI_ACPI_6_0_IO_REMAPPING_NODE *)((UINT8 *)Iort + Iort->NodeOffset);
+
+  for (Count = 0; Count < Iort->NumNodes; Count++) {
+    if ((Node->Type == EFI_ACPI_IORT_TYPE_ROOT_COMPLEX) || (Node->Type == EFI_ACPI_IORT_TYPE_NAMED_COMP)) {
+      ZeroMem (&StreamEntryConfig, sizeof (SMMU_STREAM_ENTRY_CONFIG));
```
Raymond-MS: Is there a reason to zero the StreamEntryConfig and initialize it below if Node->NumIdMappings is not > 0? Should this be pushed down into that if and/or, like in my previous comment, would it be better to combine that if into the one above?

Raymond-MS (approved):
LGTM - Just some minor comments. :)

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: If we have these functions fail during exit boot services, we have big problems because the SMMU register writes failed. So the communication with the smmu would be broken, so we assert and prevent further execution.

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: this is ok, it is just freeing some memory, the critical path of execution is within the IoMmu Protocol. The error status will be returned for SmmuDxe so the user will know what went wrong pretty clearly during the SmmuDxe driver execution.

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: mistake, fixed to Status

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: We want to zero out every StreamEntryConfig for every stream id. Even though this case would be for where there is a unused stream id, we still want to make sure we dont write bad data to the stream table entries for any stream id.

Raymond-MS (approved):

PR #355, opened 2025-03-20
[Moved to new branch] ArmPkg: Add changes required for AARCH64 MSVC

VivianNK:
`##` Description

Moved to https://github.com/microsoft/mu_silicon_arm_tiano/pull/368
Add changes required for AARCH64 MSVC

For details on how to complete these options and their meaning refer to [CONTRIBUTING.md](https://github.com/microsoft/mu/blob/HEAD/CONTRIBUTING.md).

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested

Built in project consuming ArmGicV3Dxe and ArmMonitorLib

## Integration Instructions

N/A

PR #368, opened 2025-04-18
[REBASE & FF] ArmPkg: Add changes required for AARCH64 MSVC

VivianNK:
`##` Description

Add changes required for AARCH64 MSVC (Moved from old PR: https://github.com/microsoft/mu_silicon_arm_tiano/pull/355)
- Add missing functions in ArmGicV3.masm and include the .masm file in ArmGicDxe infs. 
- ArmMonitorLib.masm: change ALIGN value to fix compiler error `Compiler #2070 improper alignment value 3; expected [1,2,4,8...32768]`
- Add UINT32 ControlSystemRegValue to fix compiler type conversion warning. Previously, RegValue was being used for two values, one UINT32 and one UINT64.
- Make AllocatePageTableMemory static in CpuDxe. This fixes linker and compiler errors with ArmMmuLibPageTableAlloc. The MSVC errors were LNK2005, LNK1169, and Compiler #4744

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built in project consuming ArmGicV3Dxe and ArmMonitorLib

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/ArmGicDxe/GicV3/AArch64/ArmGicV3.masm
+++ b/ArmPkg/Drivers/ArmGicDxe/GicV3/AArch64/ArmGicV3.masm
@@ -23,6 +23,8 @@
     EXPORT ArmGicV3AcknowledgeInterrupt
     EXPORT ArmGicV3SetPriorityMask
     EXPORT ArmGicV3SetBinaryPointer
+    EXPORT ArmGicV3SetControlRegister
+    EXPORT ArmGicV3GetControlRegister
 
 //UINT32
 //EFIAPI
@@ -118,5 +120,25 @@ ArmGicV3SetBinaryPointer PROC
         msr     ICC_BPR1_EL1, x0
         ret
 ArmGicV3SetBinaryPointer ENDP
+
+//UINTN
+//EFIAPI
+//ArmGicV3GetControlRegister (
+//  VOID
+//  );
+ArmGicV3GetControlRegister PROC
+        mrs     x0, ICC_CTLR_EL1
+        ret
+ArmGicV3GetControlRegister ENDP
+
+//VOID
+//EFIAPI
+//ArmGicV3SetControlRegister (
+//  IN UINTN         Value
+//  );
+ArmGicV3SetControlRegister PROC
+        msr     ICC_CTLR_EL1, x0
+        ret
+ArmGicV3SetControlRegister ENDP
 
     END
--- a/ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
+++ b/ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
@@ -605,6 +605,7 @@ GicV3DxeInitialize (
   UINT64      MpId;
   UINT64      CpuTarget;
   UINT64      RegValue;
+  UINT32      ControlSystemRegValue; // MU_CHANGE - Add ControlSystemRegValue
 
   // Make sure the Interrupt Controller Protocol is not already installed in
   // the system.
@@ -626,9 +627,11 @@ GicV3DxeInitialize (
   mGicRedistributorBase = GicGetCpuRedistributorBase (PcdGet64 (PcdGicRedistributorsBase));
   mGicNumInterrupts     = ArmGicGetMaxNumInterrupts (mGicDistributorBase);
 
-  RegValue = ArmGicV3GetControlSystemRegisterEnable ();
-  if ((RegValue & ICC_SRE_EL2_SRE) == 0) {
-    ArmGicV3SetControlSystemRegisterEnable (RegValue | ICC_SRE_EL2_SRE);
+  // MU_CHANGE [BEGIN] - Add ControlSystemRegValue
+  ControlSystemRegValue = ArmGicV3GetControlSystemRegisterEnable ();
+  if ((ControlSystemRegValue & ICC_SRE_EL2_SRE) == 0) {
+    ArmGicV3SetControlSystemRegisterEnable (ControlSystemRegValue | ICC_SRE_EL2_SRE);
+    // MU_CHANGE [END] - Add ControlSystemRegValue
     ASSERT ((ArmGicV3GetControlSystemRegisterEnable () & ICC_SRE_EL2_SRE) != 0);
   }
 
--- a/ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
+++ b/ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
@@ -104,6 +104,7 @@ FindPoolToAllocateFrom (
   @retval A pointer to the allocated page table memory or NULL if the
           allocation failed.
 **/
+STATIC
 VOID *
 AllocatePageTableMemory (
   IN UINTN  Pages
@@ -152,7 +153,7 @@ AllocatePageTableMemory (
   return Buffer;
 }
 
-PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
+STATIC PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
   AllocatePageTableMemory
 };
 
--- a/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
+++ b/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
@@ -29,6 +29,7 @@
 
 [Sources.AARCH64]
   GicV3/AArch64/ArmGicV3.S
+  GicV3/AArch64/ArmGicV3.masm  | MSFT
 
 [Packages]
   MdePkg/MdePkg.dec
--- a/ArmPkg/Drivers/ArmGicDxe/ArmGicV3Dxe.inf
+++ b/ArmPkg/Drivers/ArmGicDxe/ArmGicV3Dxe.inf
@@ -27,6 +27,7 @@
 
 [Sources.AARCH64]
   GicV3/AArch64/ArmGicV3.S
+  GicV3/AArch64/ArmGicV3.masm  | MSFT
 
 [Packages]
   MdePkg/MdePkg.dec
--- a/ArmPkg/Library/ArmMonitorLib/AArch64/ArmMonitorLib.masm
+++ b/ArmPkg/Library/ArmMonitorLib/AArch64/ArmMonitorLib.masm
@@ -6,8 +6,7 @@
 //
 //
 
-    AREA |.text|, CODE, READONLY
-    ALIGN 3
+    AREA |.text|,ALIGN=3,CODE, READONLY
 
     EXPORT ArmMonitorCall
 
--- a/ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.masm
+++ b/ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.masm
@@ -5,7 +5,7 @@
 //
 //------------------------------------------------------------------------------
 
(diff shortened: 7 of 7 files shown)


ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
```diff
@@ -626,9 +627,10 @@ GicV3DxeInitialize (
   mGicRedistributorBase = GicGetCpuRedistributorBase (PcdGet64 (PcdGicRedistributorsBase));
   mGicNumInterrupts     = ArmGicGetMaxNumInterrupts (mGicDistributorBase);
 
-  RegValue = ArmGicV3GetControlSystemRegisterEnable ();
-  if ((RegValue & ICC_SRE_EL2_SRE) == 0) {
-    ArmGicV3SetControlSystemRegisterEnable (RegValue | ICC_SRE_EL2_SRE);
+  GicV3RegValue   = ArmGicV3GetControlSystemRegisterEnable ();
+  GicV3RegValue32 = (UINT32)GicV3RegValue;
+  if ((GicV3RegValue & ICC_SRE_EL2_SRE) == 0) {
+    ArmGicV3SetControlSystemRegisterEnable (GicV3RegValue32 | ICC_SRE_EL2_SRE);
```
os-d: Why not only use `GicV3RegValue32` here? And naming it to `ControlSystemReg` or something more accurate.

```c
//UINT32
//EFIAPI
//ArmGicV3GetControlSystemRegisterEnable (
//  VOID
//  );

//VOID
//EFIAPI
//ArmGicV3SetControlSystemRegisterEnable (
//  IN UINT32         ControlSystemRegisterEnable
//  );
```

One function returns a UINT32 and the other takes a UINT32, right?

ArmPkg/Drivers/ArmGicDxe/GicV3/AArch64/ArmGicV3.masm
```diff
@@ -119,4 +121,24 @@ ArmGicV3SetBinaryPointer PROC
...
+//  );
+ArmGicV3GetControlRegister PROC
+        mrs     x0, ICC_CTLR_EL1
+        ret
+ArmGicV3GetControlRegister ENDP
+
+//VOID
+//EFIAPI
+//ArmGicV3SetControlRegister (
+//  IN UINT32         Value
```
os-d: This is incorrect, right? It should take in a `UINTN`: https://github.com/tianocore/edk2/blob/5e5ca20bf772bc80ca71b996def6f8ebad2860bf/ArmPkg/Drivers/ArmGicDxe/GicV3/Arm/ArmGicV3.S#L90-L92

https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/ICC-CTLR-EL1--Interrupt-Controller-Control-Register--EL1-

ArmPkg/Drivers/ArmGicDxe/GicV3/AArch64/ArmGicV3.masm
```diff
@@ -119,4 +121,24 @@ ArmGicV3SetBinaryPointer PROC
         ret
 ArmGicV3SetBinaryPointer ENDP
 
+//UINT32
+//EFIAPI
+//ArmGicV3GetControlRegister (
+//  VOID
+//  );
```
os-d, line 128: This is incorrect, right? Should return a UINTN? 

https://github.com/tianocore/edk2/blob/5e5ca20bf772bc80ca71b996def6f8ebad2860bf/ArmPkg/Drivers/ArmGicDxe/GicV3/Arm/ArmGicV3.S#L82-L85

https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/ICC-CTLR-EL1--Interrupt-Controller-Control-Register--EL1-

ArmPkg/Include/Protocol/ArmPageTableMemoryAllocation.h
```diff
@@ -34,7 +34,7 @@ VOID *
   );
 
 typedef struct _PAGE_TABLE_MEM_ALLOC_PROTOCOL {
-  ALLOCATE_PAGE_TABLE_MEMORY    AllocatePageTableMemory;
+  ALLOCATE_PAGE_TABLE_MEMORY    AllocatePageTableMem;
```
os-d: Can you please separate the individual changes out into separate commits and REBASE & FF this? One major problem we've had with AARCH64 MSVC changes before is they are all merged together and become unwieldy and hard to deal with.

os-d: Also, was renaming the field really necessary? I assume the name conflict was not for this but the function itself.

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
VivianNK: Updated

ArmPkg/Include/Protocol/ArmPageTableMemoryAllocation.h
VivianNK: Separated logic into 5 commits - let me know what you think

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -22,20 +22,20 @@ PAGE_TABLE_MEM_ALLOC_PROTOCOL  *mPageTableMemAllocProtocol = NULL;
   @return A pointer to the allocated buffer or NULL if allocation fails
 **/
 VOID *
-AllocatePageTableMemory (
+MmuAllocatePageTableMemory (
```
VivianNK: @os-d @kuqin12 Is this function supposed to be consumed in ArmMmuLibCore.c? Because that file has a the function declaration for `AllocatePageTableMemory` but it's not clear what its source is supposed to be. 
If this function's name is `AllocatePageTableMemory` there is a linking error: 

> error LNK2005: AllocatePageTableMemory already defined in ArmCpuDxe.lib(PageTableMemoryAllocation.obj)
INFO - ArmCpuDxe.dll.wtemp : fatal error LNK1169: one or more multiply defined symbols found

I previously made the function static but that prevents it from being consumed since it's not published as a protocol.

kuqin12: This seems like a naming conflict, did the log say which other function you are clashing with?

os-d: You should be able to just make the one in CpuDxe `STATIC`, right? It looks like that is what the conflict is. The CpuDxe one should not be directly called outside of the file, only through the protocol

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
```diff
@@ -628,7 +628,7 @@ GicV3DxeInitialize (
 
   RegValue = ArmGicV3GetControlSystemRegisterEnable ();
   if ((RegValue & ICC_SRE_EL2_SRE) == 0) {
-    ArmGicV3SetControlSystemRegisterEnable (RegValue | ICC_SRE_EL2_SRE);
+    ArmGicV3SetControlSystemRegisterEnable ((UINT32)RegValue | ICC_SRE_EL2_SRE);
```
os-d: This isn't needed now, right, because the next commit changes it?

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
```diff
@@ -688,7 +688,7 @@ GicV3DxeInitialize (
   ArmGicV3SetPriorityMask (0xff);
 
   // Use combined priority drop and deactivate (EOImode == 0)
-  RegValue  = ArmGicV3GetControlRegister ();
+  RegValue  = (UINT64)ArmGicV3GetControlRegister ();
```
os-d: Is this cast needed? Presumably GCC doesn't need it, if it is already checked in, and we only do MSVC for ARM64 where UINTN is UINT64, so does it actually complain without this?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
VivianNK: Making the function static in CpuDxe worked to fix that conflict.

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
VivianNK: dropped

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
VivianNK: dropped

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
```diff
@@ -12,7 +12,7 @@
 #include <Library/MemoryAllocationLib.h>
 #include <Protocol/ArmPageTableMemoryAllocation.h>
 
-PAGE_TABLE_MEM_ALLOC_PROTOCOL  *mPageTableMemAllocProtocol = NULL;
+PAGE_TABLE_MEM_ALLOC_PROTOCOL  *mArmMmuPageTableMemAllocProtocol = NULL;
```
os-d: What does this conflict with? If you change this back, does it still give an error?

VivianNK: It conflicts with the variable in 
`armpkg\drivers\cpudxe\pagetablememoryallocation.c`:
The error is that here the variable is a pointer type but in CpuDxe it's just the struct type.
`Compiler #4744 from  'mPageTableMemAllocProtocol' has different type`

os-d: These can be made static right, instead of changing the name?

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -152,7 +153,7 @@ AllocatePageTableMemory (
   return Buffer;
 }
 
-PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
+static PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
```
VivianNK, line 156: @os-d We do lowercase static for variables, right?

ArmPkg/Library/ArmMmuLib/AArch64/ArmMmuLibPageTableAlloc.c
VivianNK: That worked, yes. Updated the commit

os-d: We should continue to use `STATIC`. We only used lowercase `static` for a local variable that is marked static.

os-d (approved):

kuqin12 (approved):

apop5 (approved):

PR #372, opened 2025-05-02
[REBASE & FF] Backporting arm fixes to release branch

VivianNK:
## Description

Backport from https://github.com/microsoft/mu_silicon_arm_tiano/pull/368

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested

Built in project consuming ArmGicV3Dxe and ArmMonitorLib

## Integration Instructions

N/A

PR #373, opened 2025-05-05
[REBASE & FF] Minor fixes to squash on rebase

VivianNK:
## Description

Formatting fixes for previous PR: https://github.com/microsoft/mu_silicon_arm_tiano/pull/368
Change STATIC to static
Add GCC tags in inf

Waiting on manual backport of original changes into release/202502 https://github.com/microsoft/mu_silicon_arm_tiano/pull/372

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A
## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
+++ b/ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
@@ -104,7 +104,7 @@ FindPoolToAllocateFrom (
   @retval A pointer to the allocated page table memory or NULL if the
           allocation failed.
 **/
-STATIC
+static
 VOID *
 AllocatePageTableMemory (
   IN UINTN  Pages
@@ -153,7 +153,7 @@ AllocatePageTableMemory (
   return Buffer;
 }
 
-STATIC PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
+static PAGE_TABLE_MEM_ALLOC_PROTOCOL  mPageTableMemAllocProtocol = {
   AllocatePageTableMemory
 };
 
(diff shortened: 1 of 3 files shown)


ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
```diff
@@ -104,7 +104,7 @@ FindPoolToAllocateFrom (
   @retval A pointer to the allocated page table memory or NULL if the
           allocation failed.
 **/
-STATIC
+static
```
os-d, line 107: The style guidelines have been updated to use `static` over `STATIC` (though don't disallow it), but I don't think edk2 PRs have actually switched to this. I have to go back and see what the GoogleTest decision was here, but I don't quite recall or really am too worried about it if we want to go this direction.

os-d (approved):

apop5 (approved):

ArmPkg/Drivers/CpuDxe/PageTableMemoryAllocation.c
VivianNK: https://edk2.groups.io/g/devel/topic/110977230#msg121076

os-d: That’s what I’m saying that the style guidelines have been updated but I don’t think edk2 has been following that (yet). Certainly no codebase wide PR has happened for it. And changing to lowercase static will break GoogleTest (which isn’t being used here) and I remember talking on the mailing list about some schemes to fix it, but I don’t recall anything being implemented. But maybe I did. More just mentioning that I will probably still following STATIC until edk2 tells us otherwise in PR review, but I don’t care too much, just noting it might cause an issue if we were ever to write a GoogleTest for this, which we probably won’t.

PR #374, opened 2025-05-19
Add RMR support for IORT and SmmuDxe IoMmu mappings

eeshanl:
## Description

Add RMR support for IORT and SmmuDxe IoMmu mappings
Parses IORT for RMR node memory range descriptors
Maps them to the SMMU page tables to be used by the IoMmu protocol

Updates TLIBI operation from TLIBI ALL to TLIBI S2 IPA, now invalidates TLB for InputAddress to Stage 2 SMMU.

Due to an issue seen while testing on a physical arm platform, we see a 0x13 F_PERMISSION fault from the smmu when not setting both r/w bits. For now, temporarily setting both RW bits in  IoMmuSetAttribute
TODO: https://github.com/microsoft/mu_silicon_arm_tiano/issues/375 Debug this issue and revert this commit

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU SBSA and physical arm platform

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -109,6 +109,7 @@ UpdateMapping (
   UINT32      Index;
   PAGE_TABLE  *Current;
   UINT64      Entry;
+  UINT32      SmmuIndex;
 
   // Flags must be 12 bits or less
   if ((Root == NULL) || ((Flags & ~PAGE_TABLE_BLOCK_OFFSET) != 0) || (PhysicalAddress == 0)) {
@@ -153,7 +154,7 @@ UpdateMapping (
       if (Valid) {
         Entry = (PhysicalAddress & ~PAGE_TABLE_BLOCK_OFFSET); // Assign PA
         // validate entry and set leaf level flags
-        Entry                  |= PAGE_TABLE_ACCESS_FLAG | PAGE_TABLE_DESCRIPTOR | PAGE_TABLE_ENTRY_VALID_BIT;
+        Entry                  |= Flags | PAGE_TABLE_ACCESS_FLAG | PAGE_TABLE_DESCRIPTOR | PAGE_TABLE_ENTRY_VALID_BIT;
         Current->Entries[Index] =  Entry;
       } else {
         Current->Entries[Index] = Current->Entries[Index] & ~PAGE_TABLE_ENTRY_VALID_BIT; // only invalidate leaf entry
@@ -166,7 +167,24 @@ UpdateMapping (
     }
   }
 
-  ArmDataSynchronizationBarrier ();
+  // Invalidate TLBI Command.
+  // Current implementation of IoMmu protocol doesnt distinguish between stream id's or Smmu's.
+  // As a result we have to invalidate TLB for all SMMU's for the given virtual address to make
+  // sure that the correct smmu's streamid's page table's TLB is invalidated.
+  // To workaround the limitations of the IoMmu protocol, we globally set the page table root
+  // for all SMMU's and stream id's to be the same.
+  // Todo: Update the IoMmu protocol to support multiple SMMU's and stream id's so we don't have to
+  // invalidate all SMMU's TLB for a given virtual address, just the one that was updated.
+  if (!Valid) {
+    for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+      Status = SmmuV3TLBInvalidateAddress (&mIoMmu->SmmuInfo[SmmuIndex], VirtualAddress);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB\n", __func__));
+        goto End;
+      }
+    }
+  }
+
   SpeculationBarrier ();
 
 End:
@@ -188,7 +206,6 @@ End:
   @retval EFI_INVALID_PARAMETER  Invalid parameter.
   @retval EFI_OUT_OF_RESOURCES   Out of resources.
 **/
-STATIC
 EFI_STATUS
 UpdatePageTable (
   IN PAGE_TABLE  *Root,
@@ -202,7 +219,6 @@ UpdatePageTable (
   EFI_STATUS            Status;
   EFI_PHYSICAL_ADDRESS  PhysicalAddressEnd;
   EFI_PHYSICAL_ADDRESS  CurPhysicalAddress;
-  UINT32                SmmuIndex;
 
   if ((Root == NULL) || ((Flags & ~PAGE_TABLE_BLOCK_OFFSET) != 0) || (PhysicalAddress == 0) || (Bytes == 0)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
@@ -221,17 +237,6 @@ UpdatePageTable (
     }
 
     CurPhysicalAddress += EFI_PAGE_SIZE;
-  }
-
-  // Invalidate TLBI Command
-  if (!Valid) {
-    for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
-      Status = SmmuV3TLBInvalidateAll (&mIoMmu->SmmuInfo[SmmuIndex]);
-      if (EFI_ERROR (Status)) {
-        DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB.\n", __func__));
-        goto End;
-      }
-    }
   }
 
 End:
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -497,7 +497,7 @@ IoMmuSetAttribute (
              mIoMmu->SmmuInfo->PageTableRoot,
              MapInfo->PhysicalAddress,
              MapInfo->NumberOfBytes,
-             PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS (IoMmuAccess),
+             PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS ((EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE)), // TODO: https://github.com/microsoft/mu_silicon_arm_tiano/issues/375 debug issue on physical platform and revert the permissions
              FALSE,
              TRUE
              );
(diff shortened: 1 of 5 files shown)


ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -492,7 +491,7 @@ IoMmuSetAttribute (
              mIoMmu->SmmuInfo->PageTableRoot,
              MapInfo->PhysicalAddress,
              MapInfo->NumberOfBytes,
-             PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS (IoMmuAccess),
+             PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS ((EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE)), // TODO: debug issue on physical platform and revert the permissions
```
cfernald: Do you intend to revert this before merging this PR?

ArmPkg/Drivers/SmmuDxe/IoMmu.h
```diff
@@ -23,6 +23,39 @@
 #define PAGE_TABLE_DESCRIPTOR       (0x1 << 1)
 #define PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS(IoMmuAccess)  (IoMmuAccess << 6)
 
+typedef UINT64 PAGE_TABLE_ENTRY;
+
+#define PAGE_TABLE_SIZE  EFI_PAGE_SIZE / sizeof(PAGE_TABLE_ENTRY)  // Number of entries in a page table
```
cfernald:
```suggestion
#define PAGE_TABLE_SIZE  (EFI_PAGE_SIZE / sizeof(PAGE_TABLE_ENTRY))  // Number of entries in a page table
```

I find that math based C macros can be dangerous without () as that can cause unexpected order-of-operation issues if inlined with other math. Good practice to just always wrap them in ().

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -350,6 +353,28 @@ SmmuV3BuildStreamTableEntry (
...
+    for (NumMemRangeDesc = 0; NumMemRangeDesc < RmrNode->NumMemRangeDesc; NumMemRangeDesc++) {
+      IortMemRangeDesc = (EFI_ACPI_6_0_IO_REMAPPING_MEM_RANGE_DESC *)((UINT8 *)RmrNode + RmrNode->MemRangeDescRef);
+      if ((IortMemRangeDesc[NumMemRangeDesc].Base > 0) && (IortMemRangeDesc[NumMemRangeDesc].Length > 0)) {
+        Status = UpdatePageTable (SmmuInfo->PageTableRoot, IortMemRangeDesc[NumMemRangeDesc].Base, IortMemRangeDesc[NumMemRangeDesc].Length, 0, TRUE, FALSE);
+        if (EFI_ERROR (Status)) {
+          DEBUG ((DEBUG_ERROR, "%a: Failed to update RMR mapping.\n", __func__));
+          return Status;
+        }
+
+        Status = UpdatePageTable (SmmuInfo->PageTableRoot, IortMemRangeDesc[NumMemRangeDesc].Base, IortMemRangeDesc[NumMemRangeDesc].Length, PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS (3), FALSE, TRUE);
```
cfernald: Is there a reason this has to be done in parts? The creation then setting of attributes vs just setting the attributes initially?

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.h
```diff
@@ -23,6 +23,39 @@
...
+  @param [in]  Flags                      Flags to set for the mapping. 12 bits or less.
+  @param [in]  Valid                      Boolean to indicate if the entry is valid.
+  @param [in]  SetReadWriteFlagsOnly      Boolean to indicate if only R/W flags should be set.
+
+  @retval EFI_SUCCESS            Success.
+  @retval EFI_INVALID_PARAMETER  Invalid parameter.
+  @retval EFI_OUT_OF_RESOURCES   Out of resources.
+**/
+EFI_STATUS
+UpdatePageTable (
```
os-d, line 50: Doesn't need to block here, but still wondering if the page table management logic can be shared with the MMU page table logic, at least in part.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -1152,7 +1152,12 @@ SmmuV3GetStreamIdInfo (
 
               // Store the Stream ID range information
               for (CurID = StartId; CurID <= EndId; CurID++) {
-                CopyMem (&SmmuInfoArray[SmmuIndex].StreamEntryConfig[CurID], &StreamEntryConfig, sizeof (SMMU_STREAM_ENTRY_CONFIG));
+                if (Node->Type == EFI_ACPI_IORT_TYPE_RMR) {
+                  SmmuInfoArray[SmmuIndex].StreamEntryConfig[CurID].RmrNode = (EFI_ACPI_6_0_IO_REMAPPING_RMR_NODE *)Node;
+                } else {
+                  SmmuInfoArray[SmmuIndex].StreamEntryConfig[CurID].CacheCoherentAttribute = StreamEntryConfig.CacheCoherentAttribute;
+                  SmmuInfoArray[SmmuIndex].StreamEntryConfig[CurID].MemoryAccessFlags      = StreamEntryConfig.MemoryAccessFlags;
```
os-d, line 1204: I didn't walk through the whole logic, but the whole copy wasn't needed, just these two fields (or these fields make up the whole entry?).

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: @cfernald Seeing an issue on a physical arm device where If i just pass in IoMmuAccess, which is passed in from the caller of SetAttribute, ie PciIo, causes a PERMISIONS 0x13 fault on the smmu for the nvme. When i set the permissions to be both R and W this issue is no longer seen. Originally, on physical platforms, i would just noop the SetAttribute function and set bot RW in IoMmuMap. However, to have this code be properly compatible with all platforms, wanted to set both R/W for now until I figure out what is causing the PERMISIONS fault.

ArmPkg/Drivers/SmmuDxe/IoMmu.h
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: you are right, the way I wrote this originally was to mimic the way IoMmuMap and SetAttribute get called in PciIo, but there is no need to do that here. Fixed to mapping and setting attributes in one go.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: @os-d StreamEntryConfig has 3 members 
- RmrNode 
- CacheCoherentAttribute 
- MemoryAccessFlags

The Rmr node pointer is only needed for stream id's that have a RMR mem range descriptor defined. RMR nodes do not have a CCA or MAF fields. However, Named Component, and RC nodes do, and they don't necessarily have to have a RMR node associated with them.
The Stream ID for the RMR range coincides to a stream id within an id mapping range of a RC/Named Component node.

As a result, we only set the appropriate members depending on the current node/stream id we are iterating over in the IORT.

ArmPkg/Drivers/SmmuDxe/IoMmu.h
eeshanl: When I first implemented this, I looked at the MMU page logic to see if I could re-use, at an initial glance it seemed like there may be some issues with re-using the code, so I had decided to implement it myself. I will make a task for myself to look into implementing/sharing the SMMU page table logic with the MMU logic.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: Alright, maybe file an issue to track this to make sure it gets cleaned up once that is understood.

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: https://github.com/microsoft/mu_silicon_arm_tiano/issues/375

ArmPkg/Drivers/SmmuDxe/IoMmu.h
eeshanl: https://github.com/microsoft/mu_silicon_arm_tiano/issues/376

ArmPkg/Drivers/SmmuDxe/IoMmu.c
os-d: I'm a bit confused on this issue, basically the mapping code is trying to set RO or WO(?) and that is causing failures on physical platforms, so we are just always setting RW? I'm confused on what you were doing before with the no-op.

Basically, I'm trying to understand what the impact of checking in this workaround is: are we degrading the security status by giving more permissions to pages than we should be?

eeshanl: @os-d Yes, this would temporally loosen the R/W permissions.

On the physical platform, When I set RO or WO that gets passed into SetAttribute(), or directly in Map(), I see these PERMISIONS faults. In my testing, I also tried to noop the SetAttribute() function, and directly set the R/W permissions that are passed in inside Map(), but this also led to the same PERMISIONS faults.

The behavior I see is, during the NVME dma process, it seems to try to Write while RO is set or Read while WO is set. Setting it to both R and W resolved this issue.

eeshanl: After some more debugging and testing of this issue, seems like if I go back to my original method of nooping the SetAttribute() function, and appropriately setting RO, WO, or RW in IoMmuMap() itself, seems to do the trick. With the existing implementation, SetAttribute would have to then later go and update the R/W bits. I think the last time i tested this exact case was before my concatenated table bug fix, so I think I incorrectly assumed it would still not work.

I will now use the Iommu operation that is passed in to the IoMmuMap function to directly map to the page table R/W bits and noop the SetAttribute function. I think that is more appropriate. As if you look in PciIo, the Operation is converted to IoMmuAccess R/W bits the same way as I plan to anyway within IoMmuMap.
Tldr for SMMU, SetAttribute is highly redundant, and I do not see the point of it.

os-d (approved):
As noted in the comments, I'm okay with this going through while debugging the issue that requires both R and W permissions to be set, but that that should be addressed.

PR #385, opened 2025-06-06
Update to allow platform to enable/disable SMMUs

eeshanl:
## Description
Update to allow platform to enable/disable SMMUs

Allows the platform to toggle translation/bypass (enable/disable) on any SMMU within the IORT during pre-boot.
The SMMU_CONFIG HOB structure is updated to allow the platform to configure a SMMU_STATUS structure within SMMU_CONFIG to enable/disable any SMMU within the IORT.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on qemu sbsa and physical arm device

## Integration Instructions

Same as before, pass in SMMU_CONFIG HOB structure to SmmuDxe

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
+++ b/ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
@@ -992,13 +992,13 @@ SmmuV3ExitBootServices (
   for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
     Status = SmmuV3DisableTranslation (mIoMmu->SmmuInfo[SmmuIndex].SmmuBase);
     if (EFI_ERROR (Status)) {
-      DEBUG ((DEBUG_ERROR, "%a: Failed to disable smmu translation.\n", __func__));
+      DEBUG ((DEBUG_ERROR, "%a: Failed to disable smmu 0x%llx translation.\n", __func__, mIoMmu->SmmuInfo[SmmuIndex].SmmuBase));
       ASSERT_EFI_ERROR (Status);
     }
 
     Status = SmmuV3SetGlobalBypass (mIoMmu->SmmuInfo[SmmuIndex].SmmuBase);
     if (EFI_ERROR (Status)) {
-      DEBUG ((DEBUG_ERROR, "%a: Failed to set global bypass.\n", __func__));
+      DEBUG ((DEBUG_ERROR, "%a: Failed to set smmu 0x%llx global bypass.\n", __func__, mIoMmu->SmmuInfo[SmmuIndex].SmmuBase));
       ASSERT_EFI_ERROR (Status);
     }
   }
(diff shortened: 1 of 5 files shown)


kuqin12 (dismissed):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -1110,12 +1111,43 @@ InitializeSmmuDxe (
     goto Error;
   }
 
+  // Set SMMUs' Enabled status based on the SMMU_STATUS in the SMMU_CONFIG HOB structure.
+  for (SmmuStatusIndex = 0; SmmuStatusIndex < SmmuConfig->SmmuCount; SmmuStatusIndex++) {
```
cfernald, line 1115: From the description it implies that "enabled = false" => bypass mode but more accurately it will just be left whatever configuration it starts as, right?

In this case, is there a benefit to track this and have to condition all future accesses vs just not putting it in the global list/count?

eeshanl: @cfernald 

So the way I have modified the logic is
1. Find Enabled Status for all smmu's and update global struct.
2. configure only the smmu's that are found enabled
3. Disable and bypass all smmu's that are disabled afterwards at end of driver

So no, this would not leave the smmu in whatever state it was found in, as it explicitly disables and sets to bypass at the end if disabled by the platform hob config.

cfernald (approved):

Raymond-MS (approved):
LGTM - However, I only see this new enable field you added being updated by the mIoMmu enable which is what is being used mainly throughout this PR. Where is the SmmuConfig enable meant to be used? I assume in a different set of files?

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -1110,12 +1111,43 @@ InitializeSmmuDxe (
     goto Error;
   }
 
+  // Set SMMUs' Enabled status based on the SMMU_STATUS in the SMMU_CONFIG HOB structure.
+  for (SmmuStatusIndex = 0; SmmuStatusIndex < SmmuConfig->SmmuCount; SmmuStatusIndex++) {
+    for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+      if (mIoMmu->SmmuInfo[SmmuIndex].SmmuBase == SmmuConfig->SmmuStatus[SmmuStatusIndex].SmmuBase) {
+        mIoMmu->SmmuInfo[SmmuIndex].Enabled = SmmuConfig->SmmuStatus[SmmuStatusIndex].Enabled;
```
eeshanl, line 1118: @Raymond-MS we update the Enable value from the SmmuConfig Hob into the mIoMmu global structure.
the global struct gets used in the code throughout afterwards.

The SmmuStatus field in the SmmuConfig Hob you can essentially think of as an extension of the IORT.

Raymond-MS: I had it backwards, whoops. But yeah, that makes sense. :)

PR #388, opened 2025-06-11
Update SMMU_CONFIG HOB to use a Smmu disabled list instead.

eeshanl:
## Description

Currently, the platform will have to define explicitly each Smmu base/enabled/disabled status. This could lead to the platform accidentally setting this incorrectly.

As a result, we now set all Smmu's to enabled by default, and if a Smmu base address is found in the SmmuDisabledList within the SMMU_CONFIG HOB, then it will be disabled and set to bypass mode.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical arm device and qemu sbsa.

## Integration Instructions

Platform will create and publish this SMMU_CONFIG HOB for SmmuDxe to consume.

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/README.md
+++ b/ArmPkg/Drivers/SmmuDxe/README.md
@@ -244,27 +244,21 @@ Generic Platform Integration:
 - Append the IORT structure to this struct and update the fields accordingly.
 
   ```c
-   // SMMU_STATUS structure to hold the status of each SMMU instance.
-   // Contains the base address of the SMMU and whether it is enabled or not.
-   typedef struct _SMMU_STATUS {
-      UINT64     SmmuBase;
-      BOOLEAN    Enabled;
-   } SMMU_STATUS;
-
    // SMMU_CONFIG structure to pass the SMMU configuration data from the platform to the SMMU driver.
    typedef struct _SMMU_CONFIG {
-      UINT32         VersionMajor;
-      UINT32         VersionMinor;
-      UINT32         SmmuCount;
-      SMMU_STATUS    *SmmuStatus;
-      UINT32         IortSize;
-      UINT32         IortOffset;
+      UINT32    VersionMajor;
+      UINT32    VersionMinor;
+      UINT64    SmmuDisabledCount;                // Number of SMMU instances to be disabled.
+      UINT64    SmmuDisabledList[MAX_SMMU_COUNT]; // Array of disabled SMMU stream IDs.
+      UINT32    IortSize;
+      UINT32    IortOffset;      // Offset in bytes to the IORT table from the start of the HOB structure.
    } SMMU_CONFIG;
   ```
 
 - Essentialy the same as IORT we want to publish
 - The SMMU expects the entire IORT data to be passed into a HOB gSmmuConfigHobGuid.
 - The platform must create the IORT structure and create gSmmuConfigHobGuid with that data using BuildGuidDataHob.
+- If the platform needs to disable/bypass any Smmu, they can add the SMMU base address to the SmmuDisabledList in the HOB.
 - This structure is consumed by SmmuDxe to configure the SMMU hardware
 
 Integration with Qemu:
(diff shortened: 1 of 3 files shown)


ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -860,6 +860,11 @@ CheckSmmuConfigVersion (
     return EFI_INVALID_PARAMETER;
   }
 
+  if ((SmmuConfig->SmmuDisabledCount > 0) && (SmmuConfig->SmmuDisabledList == NULL)) {
```
cfernald: `SmmuDisabledList` is a fixed array and so should never be NULL

ArmPkg/Drivers/SmmuDxe/README.md
```diff
@@ -244,27 +244,21 @@ Generic Platform Integration:
...
-      UINT32         VersionMajor;
-      UINT32         VersionMinor;
-      UINT32         SmmuCount;
-      SMMU_STATUS    *SmmuStatus;
-      UINT32         IortSize;
-      UINT32         IortOffset;
+      UINT32    VersionMajor;
+      UINT32    VersionMinor;
+      UINT64    SmmuDisabledCount;                // Number of SMMU instances to be disabled.
+      UINT64    SmmuDisabledList[MAX_SMMU_COUNT]; // Array of disabled SMMU stream IDs.
```
cfernald: This is a list of disabled SMMU Base's, not stream IDs, from below

cfernald: Consider using the offset scheme here as well. It's pretty typical to have a header appended with variable size arrays found by offsets and counts. This way you dont have to always have MAX_SMMU_COUNT as wasted space and a limitation.

Raymond-MS (approved):
LGTM - My only nitpick comment would be that double for loop could potentially get split out into another function like "GetMatchingDisabledIndex" or something like that but in all honesty, that is minor and I wouldn't worry about it.

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: removed that check, done

ArmPkg/Drivers/SmmuDxe/README.md
eeshanl: Fixed comments

eeshanl: Changed to use 
  UINT32    SmmuDisabledListSize;   // Size of SmmuDisabledList in bytes.
  UINT32    SmmuDisabledListOffset; // Offset in bytes to the SmmuDisabledList from the start of the HOB structure.

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -851,7 +851,7 @@ GetSmmuConfigHobData (
 **/
 STATIC
 EFI_STATUS
-CheckSmmuConfigVersion (
+CheckSmmuConfigStructure (
```
apop5, line 854: It looks like after this change, the version 8 structures can no longer be supplied. Is that the intent?

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: yes, bumping the revision number up.

I am modifying/adding in:

      UINT32    SmmuDisabledListSize;   // Size of SmmuDisabledList in bytes.
      UINT32    SmmuDisabledListOffset; // Offset in bytes to the SmmuDisabledList from the start of the HOB structure.

To the HOB definition, so each platform must publish this HOB in entirety with these updated fields.

PR #398, opened 2025-07-25
ArmPkg: SmmuDxe Switch to 2-Level Stream Table Implementation

eeshanl:
## Description

Switches to 2-Level Stream Table implementation.
- If the number of StreamId's < 64: linear stream table is used. 64 is the number of STE's that can fit in EFI_PAGE_SIZE.
- Else: 2-Level implementation to consumes less memory and has less strict alignment requirements.
- Eliminates need for large contiguous allocation chunks.

Also cleans up and reduces memory usage throughout SmmuDxe.

Memory savings example:
Memory usage of StreamID space of 0xFFFFF would allocate a 64MB size and aligned linear stream table.
With 2-Level this reduces down to: 
StreamID bits Split = 6 always so leaf level table is 1 page. 2^14 entries, 16,384 entries * 8 bytes = 131,072 bytes + 4096 bytes for leaf level table = 135,168 bytes, or ~135K 

Docs:
- https://developer.arm.com/documentation/109242/0100/Programming-the-SMMU/Stream-table/2-level-Stream-table

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical arm platform.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
+++ b/ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
@@ -24,7 +24,6 @@
 #include <Library/UefiBootServicesTableLib.h>
 #include <Library/UefiDriverEntryPoint.h>
 #include <Protocol/AcpiTable.h>
-#include <Protocol/IoMmu.h>
 #include <Guid/SmmuConfig.h>
 #include "IoMmu.h"
 #include "SmmuV3.h"
@@ -295,10 +294,9 @@ SmmuV3FreeQueue (
 }
 
 /**
-  Build the stream table for SMMUv3.
+  Build the default stream table entry for SMMUv3.
 
   @param [in]  SmmuInfo       Pointer to the SMMU_INFO structure.
-  @param [in]  StreamId       Stream ID.
   @param [out] StreamEntry    Pointer to the stream table entry.
 
   @retval EFI_SUCCESS         Success.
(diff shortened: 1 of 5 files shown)


kuqin12 (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -623,8 +597,8 @@ SmmuV3Configure (
     goto End;
   }
 
-  // Allocate Linear Stream Table
-  StreamTablePtr = SmmuV3AllocateStreamTable (SmmuInfo, &StreamTableLog2Size, &StreamTableSize);
+  TwoLevelStreamTable = (SmmuInfo->StreamTableEntryMax >= (EFI_PAGE_SIZE / sizeof (SMMUV3_STREAM_TABLE_ENTRY)));
```
cfernald, line 600: My assumption is this is an optimization no not waste a second page in the case that a single page would suffice or is this explicitly not allowed. Do you know if it is common for us to have less than a page worths of entries?

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: For the linear stream table case, you can have as many Id's as you want. But to limit the allocation of the leaf level STE's (for linear case, the base table is the leaf table), we limit the linear stream table size to 1 page, else switch to 2 level.

In my testing, I have come across one SMMU where the number of streamID's was less than 64.

PR #407, opened 2025-08-26
ArmPlatformPkg/MemoryInitPeiLib: Produce V2 resource descriptor HOBs

makubacki:
## Description

Describe the cache attributes for ranges.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Boot to EFI shell on QemuSbsaPkg

## Integration Instructions

- Ensure that resource descriptor HOB consuming codes supports the V2 definition.

### DIFF
--- a/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
+++ b/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
@@ -7,6 +7,7 @@
 **/
 
 #include <PiPei.h>
+#include <Uefi.h>
 
 #include <Library/ArmMmuLib.h>
 #include <Library/ArmPlatformLib.h>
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

apop5 (approved):

os-d (approved):
My only concern is if we don't produce the v1 HOBs this will be incompatible with edk2. Some platforms pick and choose pieces between edk2 and Mu. But, I'm fine to just note that and move forward to v2.

makubacki:
> My only concern is if we don't produce the v1 HOBs this will be incompatible with edk2. Some platforms pick and choose pieces between edk2 and Mu. But, I'm fine to just note that and move forward to v2.

Yeah, I wasn't planning on upstreaming it soon for that reason. I'm also happy to produce v1 HOBs as well if we think there is a use case for consumers of this code.

makubacki:
Sounds like there is no concern. I'll go ahead and merge. It was tested with both DXE cores.

PR #431, opened 2025-10-30
Carve out the mm buffer region from the bulk system memory hob

kuqin12:
## Description

The MM communication buffer memory is being added in MM communication DXE driver through gDS calls. This means that the region should not be reported as available through resource descriptor hob.

This change will carve out the MM communication buffer from the system memory.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU SBSA and booted to OS desktop.

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
+++ b/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
@@ -103,7 +103,7 @@ MemoryPeim (
   NextHob.Raw = GetHobList ();
   while ((NextHob.Raw = GetNextHob (EFI_HOB_TYPE_RESOURCE_DESCRIPTOR, NextHob.Raw)) != NULL) {
     if ((NextHob.ResourceDescriptor->ResourceType == EFI_RESOURCE_SYSTEM_MEMORY) &&
-        (SystemMemoryBase >= NextHob.ResourceDescriptor->PhysicalStart) && // MU_CHANGE
+        (SystemMemoryBase >= NextHob.ResourceDescriptor->PhysicalStart) &&                                           // MU_CHANGE
         (NextHob.ResourceDescriptor->PhysicalStart + NextHob.ResourceDescriptor->ResourceLength <= SystemMemoryTop)) // MU_CHANGE
     {
       Found = TRUE;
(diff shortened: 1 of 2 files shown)


ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
```diff
@@ -108,14 +108,69 @@ MemoryPeim (
...
+          (PcdGet64 (PcdMmBufferBase) + PcdGet64 (PcdMmBufferSize)),
+          EFI_MEMORY_WB,
+          NULL
+          );
+      }
+    } else if ((PcdGet64 (PcdMmBufferBase) + PcdGet64 (PcdMmBufferSize)) >
+               PcdGet64 (PcdSystemMemoryBase) &&
+               (PcdGet64 (PcdMmBufferBase) + PcdGet64 (PcdMmBufferSize)) <=
+               (PcdGet64 (PcdSystemMemoryBase) + PcdGet64 (PcdSystemMemorySize))) {
+      // The end of mm communication buffer is in the system memory range
```
os-d, line 150: What about the other case where the comm buffer start is in system memory but the end is outside? That would be an odd configuration, perhaps, but should it be accounted for?

I think that maybe you could just have two conditions here:

if buffer overlaps at all:
   if conditions within for what kind of overlap
else
   report all of system memory

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
```diff
@@ -108,14 +108,69 @@ MemoryPeim (
...
-      EFI_RESOURCE_SYSTEM_MEMORY,
-      ResourceAttributes,
-      PcdGet64 (PcdSystemMemoryBase),
-      PcdGet64 (PcdSystemMemorySize),
-      EFI_MEMORY_WB,
-      NULL
-      );
+    // MU_CHANGE: Carve out MM communication buffer from system memory
+    // But pay attention to the potential overlap with the mm communication buffer
+    if (PcdGet64 (PcdMmBufferBase) >= PcdGet64 (PcdSystemMemoryBase) &&
```
os-d: There are a lot of PcdGet64 for the same value, for readability you could do those ahead of this

os-d (commented):
Does this need to be updated in PeilessSec as well?

I assume you'll be taking this to edk2 as wel.

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
kuqin12: This is covered in the first case.

The 3 cases are:
1. The start of mm buffer is in the system memory, then it will either break up the system memory into 2 regions, or shrink the system memory from the end;
2. The end of mm buffer is in the system memory, then we shrink the beginning of the system memory;
3. No overlap, report system memory as is.

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
kuqin12: Agreed, will update.

kuqin12: This is updated.

kuqin12:
> Does this need to be updated in PeilessSec as well?
>
Hmm, I think that will be needed. But I need to locate a platform to test the change. Do you happen to know one? Otherwise, I will defer this part till when trying to upstream this to EDK2.
> I assume you'll be taking this to edk2 as wel.
>
Indeed.

os-d (approved):

Raymond-MS (approved):
LGTM!

PR #444, opened 2025-11-18
ArmPkg: Improve SmmuDxe performance

eeshanl:
## Description

ArmPkg: Improve SmmuDxe performance

- Moves all mapping logic to SetAttribute() so that we don't redundantly update page table bits
- Improves scope of TPL locks in SmmuV3SendCommand() to reduce time spent waiting on lock
- Improves logic for waiting for command to complete in Command Queue

Improves hibernate/resume speeds by over 2x.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical platform with hibernate/resume to stress test a high I/O scenario.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -35,46 +35,6 @@ typedef struct IOMMU_MAP_INFO {
 } IOMMU_MAP_INFO;
 
 /**
-  Update the RW flags of a page table entry per Arm Architecture Reference Manual for A profile.
-  <https://developer.arm.com/documentation/102105/ka-07>
-
-  The bottom 12 bits of a PAGE_TABLE_ENTRY, such as R/W. Only allows setting/clearing of R/W bits
-
-  @param [in]  Table                  Pointer to the page table.
-  @param [in]  Flags                  Flags such as Read/Write Flags to set or clear. Only allows clearing of R/W bits. 12 bits or less.
-  @param [in]  Index                  Index of the entry to update. <= PAGE_TABLE_SIZE
-
-  @retval EFI_SUCCESS            Success.
-  @retval EFI_INVALID_PARAMETER  Invalid parameter.
-**/
-STATIC
-EFI_STATUS
-UpdateReadWriteFlags (
-  IN PAGE_TABLE  *Table,
-  IN UINT16      Flags,
-  IN UINT32      Index
-  )
-{
-  UINT64  Entry;
-
-  if ((Table == NULL) || ((Flags & ~PAGE_TABLE_BLOCK_OFFSET) != 0) || (Index >= PAGE_TABLE_SIZE) ||
-      ((Flags & ~(PAGE_TABLE_READ_BIT | PAGE_TABLE_WRITE_BIT)) != 0))
-  {
-    DEBUG ((DEBUG_ERROR, "%a: Invalid parameter.\n", __func__));
-    ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
-    return EFI_INVALID_PARAMETER;
-  }
-
-  // Allows clearing the R/W bits without affecting the other bits in the entry.
-  Entry = Table->Entries[Index] & ~(PAGE_TABLE_READ_BIT | PAGE_TABLE_WRITE_BIT);
-  // Set R/W bits in page table entry
-  Entry                |= Flags;
-  Table->Entries[Index] = Entry;
-
-  return EFI_SUCCESS;
-}
-
-/**
   Update the mapping of a virtual address to a physical address in the page table.
 
   Iterates through the page table levels to find the leaf entry for the given virtual address and
@@ -87,7 +47,6 @@ UpdateReadWriteFlags (
   @param [in]  PhysicalAddress            Physical address to map to.
   @param [in]  Flags                      Flags to set for the mapping. 12 bit or less.
   @param [in]  Valid                      Boolean to indicate if the entry is valid.
-  @param [in]  SetReadWriteFlagsOnly      Boolean to indicate if only flags should be set.
 
   @retval EFI_SUCCESS            Success.
   @retval EFI_INVALID_PARAMETER  Invalid parameter.
@@ -100,8 +59,7 @@ UpdateMapping (
   IN UINT64      VirtualAddress,
   IN UINT64      PhysicalAddress,
   IN UINT16      Flags,
-  IN BOOLEAN     Valid,
-  IN BOOLEAN     SetReadWriteFlagsOnly
+  IN BOOLEAN     Valid
   )
 {
   EFI_STATUS  Status;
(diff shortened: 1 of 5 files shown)


ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -748,90 +785,42 @@ SmmuV3SendCommand (
...
-  if ((Count == 0) || (ConsumerIndex != ProducerIndex)) {
-    DEBUG ((DEBUG_ERROR, "%a: Timeout waiting for command queue to be consumed\n", __func__));
-    Status = EFI_TIMEOUT;
-  }
+  // Loop until the command is consumed
+  do {
+    OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
+    SmmuV3CmdQueueUpdateCachedConsumer (SmmuInfo, QueueMask, WrapMask, TotalQueueEntries, &ConsumerIndex, &ConsumerWrap);
+    gBS->RestoreTPL (OldTpl);
+  } while (SmmuInfo->CachedConsumer < SmmuInfo->CachedProducer);
```
cfernald: Shouldn't this be compared to a local of what CachedProducer was when it was used here? It might be updated by other calls as well so this is affectively waiting for all commands to finish, not just the one sent here.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -713,6 +713,46 @@ SmmuV3WriteCommands (
...
+  UINT64            CachedConsumerWrap;
+
+  Consumer.AsUINT32 = SmmuV3ReadRegister32 (SmmuInfo->SmmuBase, SMMU_CMDQ_CONS);
+  ConsumerIndex     = Consumer.ReadIndex & QueueMask;
+  ConsumerWrap      = Consumer.ReadIndex & WrapMask;
+  *ConsumerIndexOut = ConsumerIndex;
+  *ConsumerWrapOut  = ConsumerWrap;
+
+  CachedConsumerWrap       = SmmuInfo->CachedConsumer & WrapMask;
+  SmmuInfo->CachedConsumer = SmmuInfo->CachedConsumer & ~QueueMask;
```
cfernald: nit: I would calculate the CachedConsumer value and write it as a single transaction. It shouldn't matter since this is all done at TPL_HIGH doesn't matter but simplifies reasoning about potential torn state scenarios.

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -289,30 +238,22 @@ IoMmuMap (
     goto End;
   }
 
-  PhysicalAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
-  Status          = UpdatePageTable (mIoMmu->SmmuInfo->PageTableRoot, PhysicalAddress, *NumberOfBytes, 0, TRUE, FALSE);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "%a: Failed to update page table.\n", __func__));
+  *DeviceAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress; // Identity mapping
```
os-d: Should this only be filled out if you succeed? i.e. keeping this where it was

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -510,8 +428,7 @@ IoMmuSetAttribute (
              MapInfo->PhysicalAddress,
              MapInfo->NumberOfBytes,
              PAGE_TABLE_READ_WRITE_FROM_IOMMU_ACCESS ((EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE)), // TODO: https://github.com/microsoft/mu_silicon_arm_tiano/issues/375 debug issue on physical platform and revert the permissions
```
os-d, line 430: Is the TODO no longer needed with removing the `SetReadWriteFlagsOnly` param gone?

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -713,6 +713,46 @@ SmmuV3WriteCommands (
...
+  SMMUV3_CMDQ_CONS  Consumer;
+  UINT32            ConsumerIndex;
+  UINT32            ConsumerWrap;
+  UINT64            CachedConsumerWrap;
+
+  Consumer.AsUINT32 = SmmuV3ReadRegister32 (SmmuInfo->SmmuBase, SMMU_CMDQ_CONS);
+  ConsumerIndex     = Consumer.ReadIndex & QueueMask;
+  ConsumerWrap      = Consumer.ReadIndex & WrapMask;
+  *ConsumerIndexOut = ConsumerIndex;
+  *ConsumerWrapOut  = ConsumerWrap;
```
os-d, line 755: This is a local only function, but CodeQL may get mad, so may be worth having null checks

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -748,90 +785,42 @@ SmmuV3SendCommand (
...
 
   Status = SmmuV3WriteCommands (SmmuInfo, ProducerIndex, 1, Command);
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a: Error writing command to queue\n", __func__));
-    goto End;
+    gBS->RestoreTPL (OldTpl);
+    return Status;
   }
 
   ArmDataSynchronizationBarrier ();
```
os-d: I know you didn't change this here, but a comment on why there is a dsb here would be helpful. I wonder if it should live in a different function, i.e. write commands or write reg?

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -748,90 +785,42 @@ SmmuV3SendCommand (
...
-  }
+  gBS->RestoreTPL (OldTpl);
 
-  if ((Count == 0) || (ConsumerIndex != ProducerIndex)) {
-    DEBUG ((DEBUG_ERROR, "%a: Timeout waiting for command queue to be consumed\n", __func__));
-    Status = EFI_TIMEOUT;
-  }
+  // Loop until the command is consumed
+  do {
+    OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
```
os-d, line 833: You are doing the same operation, but have two different methods of handling TPLs. Above, you have the loop entirely contained in a TPL_HIGH block and only lower TPL after you are done looping. Here you raise and lower on each iteration of the loop. I think these should either have the same behavior or a comment on why the behavior is different.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Still need to keep this TODO as the issue still exists and I need to look further into. SetReadWriteFlagsOnly is separate from this and used to be needed because we used to set the R|W bits in SetAttribute and the Valid and PhysicalAddress in Map(). Now they are all unified under SetAttribute and we set R|W, valid, and PhysicalAddress all in one go.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: Fixed

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: They do need to be handled a bit differently, added comments to explain why

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: moved to inside SmmuV3WriteCommands and added comment to explain why dsb is needed

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: fixed

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

os-d (approved):

PR #447, opened 2025-11-25
ArmPkg: Switch to TLBI All

eeshanl:
## Description

ArmPkg: Switch to TLBI All

Currently TLBI Range invalidation requires unecessary computation of Num and Scale values to construct the command.
Instead, we will simply issue a TLBI All. Has no impact on performance during high I/O scenarios.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical platform.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -146,7 +146,6 @@ UpdatePageTable (
 {
   EFI_STATUS            Status;
   EFI_PHYSICAL_ADDRESS  PhysicalAddressEnd;
-  EFI_PHYSICAL_ADDRESS  PhysicalAddressStart;
   EFI_PHYSICAL_ADDRESS  CurPhysicalAddress;
   UINT32                SmmuIndex;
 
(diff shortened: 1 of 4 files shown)


ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -180,13 +179,7 @@ UpdatePageTable (
   if (!Valid) {
     for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
       if (mIoMmu->SmmuInfo[SmmuIndex].Enabled) {
-        if (mIoMmu->SmmuInfo[SmmuIndex].RangeInvalidationSupported) {
```
Raymond-MS, line 183: I think this needs to be fully removed now, right? I don't think it was ever used outside of this case you had created previously.  I don't see the removal of the flag field from the structure.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -882,13 +882,6 @@ SmmuV3TLBInvalidateAll (
     return Status;
   }
 
-  SMMUV3_BUILD_CMD_TLBI_EL2_ALL (&Command);
```
Raymond-MS, line 885: Why was this section of code removed in relation to this PR? Feel free to resolve this comment once answered, I am just curious as it doesn't look related.

eeshanl: To issue a TLBI ALL, we were redundantly issuing a NSNH_ALL command and a TLBI_EL2_ALL. This should not be necessary as per the spec NSNH_ALL already covers invalidation of all stages, 1 and 2.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: I left the capability in there in case in the future we want to actually use range-based invalidation. During Smmu init, we will check the capability from a Smmu register and store it in this boolean here.

Raymond-MS (approved):
LGTM!

cfernald (approved):

cfernald (approved):

PR #451, opened 2025-12-08
ArmPkg: Add bounce buffering to IoMmu SmmuDxe implementation

eeshanl:
## Description

ArmPkg: Add bounce buffering to IoMmu SmmuDxe implementation

For the cases where we have:
- Non-aligned DMA buffer when we are not using CommonBuffer operations
- An address above 4G for the DMA buffer when the operation only supports below 4G

We need to bounce buffer the original DMA buffer to a newly allocated one under the above conditions. On a Read operation Map() will copy the contents of the DMA buffer into the newly allocated buffer. On a Write operation Unmap() will copy the contents of the newly allocated buffer back to the original DMA buffer.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on physical platform

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/SmmuDxe/IoMmu.c
+++ b/ArmPkg/Drivers/SmmuDxe/IoMmu.c
@@ -30,8 +30,8 @@
 **/
 typedef struct IOMMU_MAP_INFO {
   UINTN                    NumberOfBytes;
-  UINT64                   VirtualAddress;
-  UINT64                   PhysicalAddress;
+  UINT64                   DeviceAddress;
+  UINT64                   HostAddress;
   EDKII_IOMMU_OPERATION    Operation;
 } IOMMU_MAP_INFO;
 
@@ -273,20 +273,18 @@ IoMmuMap (
     }
   }
 
-  // Assert that CommonBuffer operations do not require remapping
-  ASSERT (!((NeedRemap) && ((Operation == EdkiiIoMmuOperationBusMasterCommonBuffer) || (Operation == EdkiiIoMmuOperationBusMasterCommonBuffer64))));
+  MapInfo->NumberOfBytes = *NumberOfBytes;
+  MapInfo->DeviceAddress = DmaMemoryTop;
+  MapInfo->HostAddress   = PhysicalAddress;
+  MapInfo->Operation     = Operation;
 
-  MapInfo->NumberOfBytes   = *NumberOfBytes;
-  MapInfo->VirtualAddress  = DmaMemoryTop;
-  MapInfo->PhysicalAddress = PhysicalAddress;
-  MapInfo->Operation       = Operation;
-
+  // Bounce buffer case
   if (NeedRemap) {
     Status = gBS->AllocatePages (
                     AllocateMaxAddress,
                     EfiBootServicesData,
                     EFI_SIZE_TO_PAGES (MapInfo->NumberOfBytes),
-                    &MapInfo->VirtualAddress
+                    &MapInfo->DeviceAddress
                     );
     if (EFI_ERROR (Status)) {
       FreePool (MapInfo);
@@ -301,13 +299,13 @@ IoMmuMap (
     // so the Bus Master can read the contents of the real buffer.
     //
     if ((Operation == EdkiiIoMmuOperationBusMasterRead) || (Operation == EdkiiIoMmuOperationBusMasterRead64)) {
-      CopyMem ((VOID *)(UINTN)MapInfo->VirtualAddress, (VOID *)(UINTN)MapInfo->PhysicalAddress, MapInfo->NumberOfBytes);
+      CopyMem ((VOID *)(UINTN)MapInfo->DeviceAddress, (VOID *)(UINTN)MapInfo->HostAddress, MapInfo->NumberOfBytes);
     }
   } else {
-    MapInfo->VirtualAddress = MapInfo->PhysicalAddress;
+    MapInfo->DeviceAddress = MapInfo->HostAddress;
   }
 
-  *DeviceAddress = MapInfo->VirtualAddress;
+  *DeviceAddress = MapInfo->DeviceAddress;
   *Mapping       = MapInfo;
 
 End:
@@ -343,7 +341,14 @@ IoMmuUnmap (
 
   MapInfo = (IOMMU_MAP_INFO *)Mapping;
 
-  if (MapInfo->VirtualAddress != MapInfo->PhysicalAddress) {
+  // Bounce buffer case
+  if (MapInfo->DeviceAddress != MapInfo->HostAddress) {
+    if ((MapInfo->DeviceAddress == 0) || (MapInfo->HostAddress == 0) || (MapInfo->NumberOfBytes == 0)) {
+      DEBUG ((DEBUG_ERROR, "%a: Invalid fields in MapInfo struct.\n", __func__));
+      ASSERT (FALSE);
+      return EFI_INVALID_PARAMETER;
+    }
+
     //
     // If this is a write operation from the Bus Master's point of view,
     // then copy the contents of the mapped buffer into the real buffer
@@ -351,8 +356,8 @@ IoMmuUnmap (
     //
     if ((MapInfo->Operation == EdkiiIoMmuOperationBusMasterWrite) || (MapInfo->Operation == EdkiiIoMmuOperationBusMasterWrite64)) {
       CopyMem (
-        (VOID *)(UINTN)MapInfo->PhysicalAddress,
-        (VOID *)(UINTN)MapInfo->VirtualAddress,
+        (VOID *)(UINTN)MapInfo->HostAddress,
+        (VOID *)(UINTN)MapInfo->DeviceAddress,
         MapInfo->NumberOfBytes
         );
     }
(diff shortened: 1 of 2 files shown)


ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -246,12 +251,72 @@ IoMmuMap (
...
+      // 4GB but any part of the DMA transfer being mapped is above 4GB, then
+      // map the DMA transfer to a buffer below 4GB.
+      //
+      NeedRemap    = TRUE;
+      DmaMemoryTop = SIZE_4GB - 1;
+    }
+  }
+
+  if ((NeedRemap) && ((Operation == EdkiiIoMmuOperationBusMasterCommonBuffer) || (Operation == EdkiiIoMmuOperationBusMasterCommonBuffer64))) {
+    //
```
cfernald: Is there a purpose to this block? It's conditions are the inverse of the above if statement so it cannot occur. If the purpose is to ensure internal consistency a simple assert in the remap may be cleaner.

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Thanks for the suggestion, I'll handle this with an ASSERT instead.

eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -246,12 +251,64 @@ IoMmuMap (
     goto End;
   }
 
-  *DeviceAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress; // Identity mapping
+  DmaMemoryTop    = MAX_UINTN;
+  PhysicalAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
+
+  if ((Operation != EdkiiIoMmuOperationBusMasterCommonBuffer) && (Operation != EdkiiIoMmuOperationBusMasterCommonBuffer64)) {
+    if ((*NumberOfBytes != ALIGN_VALUE (*NumberOfBytes, SIZE_4KB)) || (PhysicalAddress != ALIGN_VALUE (PhysicalAddress, SIZE_4KB))) {
```
os-d, line 258: I see that IntelVtdDxe does this as well, but why do we remap if the NumberOfBytes is not page aligned? Won't some consumers allocate a page and then only DMA to part of the page? That would still be fine, right?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -246,12 +251,64 @@ IoMmuMap (
...
+      // 4GB but any part of the DMA transfer being mapped is above 4GB, then
+      // map the DMA transfer to a buffer below 4GB.
+      //
+      NeedRemap    = TRUE;
+      DmaMemoryTop = SIZE_4GB - 1;
+    }
+  }
+
+  // Assert that CommonBuffer operations do not require remapping
+  ASSERT (!((NeedRemap) && ((Operation == EdkiiIoMmuOperationBusMasterCommonBuffer) || (Operation == EdkiiIoMmuOperationBusMasterCommonBuffer64))));
```
os-d: This shouldn't be needed because of your if guard. Fine to keep, but the if condition above is clear.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -246,12 +251,64 @@ IoMmuMap (
...
 
   MapInfo->NumberOfBytes   = *NumberOfBytes;
-  MapInfo->VirtualAddress  = *DeviceAddress;
-  MapInfo->PhysicalAddress = *DeviceAddress;
-  *Mapping                 = MapInfo;
+  MapInfo->VirtualAddress  = DmaMemoryTop;
```
os-d: This seems a little odd to says this is the virtual address, because it is not (well, it is, but not the virtual address for the physical host address). Why use this virtual address field at all? IntelVtdDxe just directly uses DeviceAddress, which seems like it would be fine?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -276,12 +333,36 @@ IoMmuUnmap (
...
+
   if ((This == NULL) || (Mapping == NULL)) {
     DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
     ASSERT (FALSE);
     return EFI_INVALID_PARAMETER;
   }
 
+  MapInfo = (IOMMU_MAP_INFO *)Mapping;
+
+  if (MapInfo->VirtualAddress != MapInfo->PhysicalAddress) {
```
os-d: I would add a comment here just to clarify that this is the bounce buffer case

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -276,12 +333,36 @@ IoMmuUnmap (
...
+        (VOID *)(UINTN)MapInfo->PhysicalAddress,
+        (VOID *)(UINTN)MapInfo->VirtualAddress,
+        MapInfo->NumberOfBytes
+        );
+    }
+
+    //
+    // Free the mapped buffer and the MAP_INFO structure.
+    //
+    gBS->FreePages (MapInfo->VirtualAddress, EFI_SIZE_TO_PAGES (MapInfo->NumberOfBytes));
```
os-d: nit: check MapInfo->VirtualAddress != NULL before freeing

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Yes, when some consumers allocate a page and then only dma part of the page, that is still OK but the consumer would have to use a CommonBuffer operation when calling Map().

When we have a CommonBuffer Operation, it is assumed per the pci protocol uefi spec that the caller has used the AllocateBuffer() to do a page allocation for the DMA buffer. So, if it is a CommonBuffer operation, it is okay to skip the alignment check.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: I think this is a variable name issue, IntelVtdDxe uses HostAddress and DeviceAddress in the MapInfo struct.

I'll switch this to use MapInfo->HostAddress MapInfo->DeviceAddress to unify and clear up the confusion.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
os-d: I'm not quite following. Why is `NumberOfBytes` required to be page aligned? I would have assumed the size of the DMA wouldn't affect whether we are remapping or not.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: removed the assert since is redundant

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Oh I see, you're specifically asking why NumberOfBytes, i.e. the end address of the DMA buffer needs to be aligned.

The reason we need the entire buffer aligned, start address and end address, is because when we map the page in the smmu page table, we need to ensure the region that's being mapped is strictly only the region that has been requested via IoMmuMap(). This way we ensure that we don't over-map a region in the page table entry.

os-d: I feel like that may cause us to overuse bounce buffers (I don’t know the callers, I would assume it is common to have NumberOfBytes be non-paged aligned). And it is worse because it is kind of a double bounce, right, where we already don’t have a common buffer and so we are copying the data twice?

Is the concern that the device will have access to the whole page, even if it should only DMA to part of it (i.e. the caller may be using part of the page for something else that the device shouldn’t access)? That would be a driver error in my opinion, they need to correctly use the page.

eeshanl: > Is the concern that the device will have access to the whole page, even if it should only DMA to part of it (i.e. the caller may be using part of the page for something else that the device shouldn’t access)? That would be a driver error in my opinion, they need to correctly use the page.

Yes, that was my concern. In my debugging and investigation, I have seen some drivers will have a struct on the stack and issue a Map() call with a Read/Write Operation and the HostAddress being the address of the struct on the stack. And sizeof(struct) for NumberOfBytes.

This would cause us to over-map the region if the end address is not aligned because the smmu page tables can only handle mappings on page boundaries.

eeshanl: Please see also discussion with Ard in comments here as well https://github.com/tianocore/edk2/pull/11656

eeshanl: Per our discussion offline, we are going to keep it as is for now and go back and change the drivers to use page allocations for DMA buffers instead. Then afterwards, we can come back and remove the end address NumberOfBytes alignment check from the bounce buffer conditions.

Also updated readme to describe bounce buffering conditions.

Created https://github.com/microsoft/mu_basecore/issues/1603 to track

os-d (approved):

PR #456, opened 2025-12-31
[Rebase & FF] Consolidates memory initialization mechanism for pre-DXE environment

kuqin12:
## Description

This change is a preparation PR for the platform to switch to PEI less based SEC model.

Specifically, this removes the previous change where the logic will try to poke holes in the main system memory. Instead, we will move to PEI less SEC phase, because the current PEI core is incapable of handling the static carveouts gracefully.

Accordingly, the MM communication driver also needs to handle the case when the system already added the MM communication buffer to GCD.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU SBSA after switching over to PEI less SEC model.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -629,13 +629,61 @@ MmCommunication2Initialize (
                   EFI_MEMORY_XP |
                   EFI_MEMORY_RUNTIME
                   );
-  if (EFI_ERROR (Status)) {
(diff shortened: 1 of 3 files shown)


os-d (approved):
Did you take the third commit to edk2? i.e. updating your earlier proposal?

kuqin12:
> Did you take the third commit to edk2? i.e. updating your earlier proposal?

Yeah, I updated the PR in EDK2 to reflect this change. I think I will leave this PR open for a bit in case the EDK2 PR has other feedback that does not align on this approach.

apop5 (approved):

PR #463, opened 2026-02-18
Update ArmSetMemoryAttributes to avoid setting invalid states during transitions

### LINKED ISSUE #454: [Bug]: ArmSetMemoryAttributes removes eXecute permission causing crash
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

During StMM initialization on platforms with MMU enabled while mapping regions from HOB_LIST we call into ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c to ArmSetMemoryAttributes to change page permissions.

This code itself is located under a page marked as RO-X. This code first sets RO permission removing eXecute property so as soon as we return to this function it causes the system to have an access permission fault and crash.

We must set both RO-X at the same time or make sure that the Execute permission is set first then the RO/RW attribute to prevent code from crashing.

### Expected Behavior

StMM should be allowed to set RO-X on the code executing the ArmSetMemoryAttributes function.

### Steps To Reproduce

1. Compile StMM which links in the ArmSetMemoryAttributes function with RO-X permission set for this memory region in StMM
2. Load StMM image onto a platform with Hafnium enabled.
3. Boot the platform and StMM crashes during initialization when setting memory attributes.

### Build Environment

```markdown
- OS(s): Hafnium latest
- Tool Chain(s): GCC5
- Targets Impacted: CIX Orion O6
```

### Version Information

```text
ea92070eb792d2817e0b069a2f2157ab5cb6cad3
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

I validated in the code if I always set RO + X and then set execute permission this resolves the issue.

     if (UseFfaAbis) {
       PermissionRequest = ARM_FFA_SET_MEM_ATTR_MAKE_PERM_REQUEST (
-                            ARM_FFA_SET_MEM_ATTR_DATA_PERM_RO,
+                            ARM_FFA_SET_MEM_ATTR_DATA_PERM_RO | ARM_FFA_SET_MEM_ATTR_CODE_PERM_X

kuqin12:
## Description

The current `ArmSetMemoryAttributes` would cycle through the input memory attributes bit by bit and could lead the memory in an invalid transient state during transitions. i.e. setting the memory to be code type (read only and executable), if the RP flag is set, it will make the memory region as data region first and then enable the executable bit, which could cause page fault on the next instruction.

This again resolves https://github.com/microsoft/mu_silicon_arm_tiano/issues/454, more properly.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on both QEMU SBSA platform and physical hardware platform.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
+++ b/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
@@ -506,7 +506,7 @@ ArmSetMemoryAttributes (
   DEBUG ((
     DEBUG_INFO,
     "%a: BaseAddress == 0x%llx, Length == 0x%llx, Attributes == 0x%llx, Mask == 0x%llx\n",
-    __FUNCTION__,
+    __func__,
     BaseAddress,
     Length,
     Attributes,
@@ -518,6 +518,14 @@ ArmSetMemoryAttributes (
   if ((Length == 0) ||
       ((NeededAttributes & ~(EFI_MEMORY_RO | EFI_MEMORY_RP | EFI_MEMORY_XP)) != 0))
   {
+    DEBUG ((DEBUG_ERROR, "%a: Length is 0 or unsupported attributes are set in Attributes.\n", __func__));
+    Status = EFI_INVALID_PARAMETER;
+    goto Done;
+  }
+
+  if ((BaseAddress % EFI_PAGE_SIZE != 0) || (Length % EFI_PAGE_SIZE != 0)) {
+    // Address and length must be aligned to page size.
+    DEBUG ((DEBUG_ERROR, "%a: Address or length is not aligned to page size.\n", __func__));
     Status = EFI_INVALID_PARAMETER;
     goto Done;
   }
@@ -561,7 +569,8 @@ ArmSetMemoryAttributes (
   while (Length > 0) {
     Status = GetMemoryPermissions (UseFfaAbis, BaseAddress, &MemoryAttributes);
     if (EFI_ERROR (Status)) {
-      break;
+      DEBUG ((DEBUG_ERROR, "%a: GetMemoryPermissions failed with Status == %r\n", __func__, Status));
+      goto Done;
     }
 
     if (Length < Size) {
@@ -576,7 +585,8 @@ ArmSetMemoryAttributes (
                  PermissionRequest
                  );
       if (EFI_ERROR (Status)) {
-        return Status;
+        DEBUG ((DEBUG_ERROR, "%a: RequestMemoryPermissionChange failed with Status == %r\n", __func__, Status));
+        goto Done;
       }
     }
 
--- a/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
+++ b/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
@@ -244,84 +244,6 @@ RequestMemoryPermissionChange (
   return SendMemoryPermissionRequest (UseFfaAbis, &SvcArgs, &Ret);
 }
 
-// MU_CHANGE: Add ArmSetMemoryRegionNoAccess function
-EFI_STATUS
-ArmSetMemoryRegionNoAccess (
-  IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
-  IN  UINT64                Length
-  )
-{
-  EFI_STATUS  Status;
-  UINT32      MemoryAttributes;
-  UINT32      PermissionRequest;
-  BOOLEAN     UseFfaAbis;
-
-  UseFfaAbis = IsFfaMemoryAbiSupported ();
-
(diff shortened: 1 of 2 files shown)


ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -596,47 +522,67 @@ ArmSetMemoryAttributes (
...
+      PermissionRequest |= ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_X << ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_SHIFT;
+    }
+  }
+
+  Size = EFI_PAGE_SIZE;
+
+  while (Length > 0) {
+    Status = GetMemoryPermissions (UseFfaAbis, BaseAddress, &MemoryAttributes);
+    if (EFI_ERROR (Status)) {
+      break;
```
cfernald: nit: there is a break here, and return below, but also have a done tag jumped to elsewhere in the function. I would be consistent with using Done if present or just make all of them return.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -596,47 +522,67 @@ ArmSetMemoryAttributes (
...
+    }
+
+    if (Length < Size) {
+      Length = Size;
```
cfernald, line 577: if BaseAddress is not aligned, would this potentially over-set attributes? That is, if BaseAddress is 0x800 and original size is 0x800, then the calls to change permissions will be 0x800 through 0x1800 because of the round-up of length.

cfernald (approved):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -596,47 +522,67 @@ ArmSetMemoryAttributes (
...
-        goto Done;
-      }
+      PermissionRequest |= ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_XN << ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_SHIFT;
     } else {
-      Status = ArmClearMemoryRegionNoExec (BaseAddress, Length);
+      PermissionRequest |= ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_X << ARM_SPM_MM_SET_MEM_ATTR_CODE_PERM_SHIFT;
+    }
+  }
+
+  Size = EFI_PAGE_SIZE;
```
Raymond-MS, line 567: nit: I see this variable being set to EFI_PAGE_SIZE only once here. Why not remove the variable and put EFI_PAGE_SIZE in place of it in those instances where it is used?

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
Raymond-MS: To add to this, it looks as though the only thing "Done" does is return the status. In those instances where we would call goto Done, we can just return the status and remove the goto entirely.

Raymond-MS (approved):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
os-d: Yeah, in other places (e.g. CpuDxe) we have checks that BaseAddress and Size are page aligned. I would consider that here. You could play some fancy games of figuring it out for them, but I think better to make the caller be explicit and know what they are doing.

os-d (approved):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
kuqin12: Removed the return in the middle of the function and stick to use goto only.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
kuqin12: Added input parameter check on these 2.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
kuqin12: I think it would be easier to change if for some reason we want a different stepping other than page size.

cfernald (approved):

PR #465, opened 2026-02-20
ArmPlatformPkg: Remove DXE Core FV placement assumption in PeilessSec

makubacki:
## Description

Instead of failing to boot if the DXE Core is not placed in the first FV instance found, decompress and search all firmware volumes that have been installed.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Boot with DXE Core placed in a compressed FV in different orders to other FVs

## Integration Instructions

- N/A - Backward compatible

### DIFF
--- a/ArmPlatformPkg/PeilessSec/PeilessSec.c
+++ b/ArmPlatformPkg/PeilessSec/PeilessSec.c
@@ -47,6 +47,35 @@ GetPlatformPpi (
   return EFI_NOT_FOUND;
 }
 
+// MU_CHANGE [BEGIN] - Remove DXE Core FV placement assumption
+
+/**
+  Decompress all discovered firmware volumes.
+
+**/
+STATIC
+VOID
+DecompressFvs (
+  VOID
+  )
+{
+  UINTN                Instance;
+  EFI_PEI_FV_HANDLE    VolumeHandle;
+  EFI_PEI_FILE_HANDLE  FileHandle;
+
+  //
+  // Decompress all firmware volume images found across all FVs.
+  //
+  for (Instance = 0; !EFI_ERROR (FfsFindNextVolume (Instance, &VolumeHandle)); Instance++) {
+    FileHandle = NULL;
+    while (!EFI_ERROR (FfsFindNextFile (EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE, VolumeHandle, &FileHandle))) {
+      FfsProcessFvFile (FileHandle, VolumeHandle);
+    }
+  }
+}
+
+// MU_CHANGE [END] - Remove DXE Core FV placement assumption
+
 /**
   SEC main routine.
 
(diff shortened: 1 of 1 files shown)


ArmPlatformPkg/PeilessSec/PeilessSec.c
```diff
@@ -47,6 +47,48 @@ GetPlatformPpi (
...
+    while (!EFI_ERROR (FfsFindNextFile (EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE, VolumeHandle, &FileHandle))) {
+      FfsProcessFvFile (FileHandle, VolumeHandle);
+    }
+  }
+
+  //
+  // Try to load the DXE Core from any available FV.
+  // Does not return on success.
+  //
+  Status = LoadDxeCoreFromFv (NULL, 0);
```
kuqin12: I get the idea of the change above, but why do we move this function inside?

ArmPlatformPkg/PeilessSec/PeilessSec.c
```diff
@@ -153,13 +195,15 @@ SecMain (
   // SEC phase needs to run library constructors by hand.
   ProcessLibraryConstructorList ();
 
-  // Assume the FV that contains the SEC (our code) also contains a compressed FV.
-  Status = DecompressFirstFv ();
```
kuqin12, line 157: Should we remove this function (the implementation)?

makubacki: It is a function in `PrePiLib` in EmbeddedPkg. Given that's a public library interface I didn't want to change it just because the call here is not used. If you think it's really not needed otherwise, I can spend more time to look up actual usage and consider it.

ArmPlatformPkg/PeilessSec/PeilessSec.c
makubacki: I thought it was simpler to see one call from `SecMain()` to DXE Core transition rather than dancing around two connected ones in the larger function. I can split it out if preferred.

os-d (approved):
I assume you'll also take this to edk2?

ArmPlatformPkg/PeilessSec/PeilessSec.c
kuqin12: No worries then. I thought this was part of this sec module.

ArmPlatformPkg/PeilessSec/PeilessSec.c
kuqin12: I think splitting would be better.

Or even move the implementation of this non-first-fv function to the PrePiLib. But we do not need to go that far in this PR.

makubacki: Okay, I'll split it out.

makubacki:
> I assume you'll also take this to edk2?

Yes

ArmPlatformPkg/PeilessSec/PeilessSec.c
makubacki: Pushed

makubacki: @kuqin12, can you please revisit and resolve this thread when you get a chance?

apop5 (approved):

kuqin12 (approved):

PR #474, opened 2026-03-11
ArmPlatformPkg: Update TpmStartupLib Status Checking in PeilessSec

Raymond-MS:
## Description

Added a check around the return status of TpmStartupInit, we don't want to assert on EFI_UNSUPPORTED. When TPM is disabled, the NULL instance of the TpmStartupLib is included which returns UNSUPPORTED.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built QemuSbsa with TPM disabled. Verified no assert occurs. 

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/PeilessSec/PeilessSec.c
+++ b/ArmPlatformPkg/PeilessSec/PeilessSec.c
@@ -186,6 +186,7 @@ SecMain (
   // MU_CHANGE [BEGIN] - Add Tpm2StartupInit call
   // Initialize the TPM before loading the DXE core
   Status = Tpm2StartupInit ();
+
   /* NOTE: EFI_UNSUPPORTED is treated as a success due to the possibility of there
    *       not being a TPM on the system and if so, the NULL instance of the startup
    *       lib should be linked in which returns UNSUPPORTED. Also, even if TPM is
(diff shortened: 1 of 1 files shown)


ArmPlatformPkg/PeilessSec/PeilessSec.c
```diff
@@ -186,7 +186,17 @@ SecMain (
   // MU_CHANGE [BEGIN] - Add Tpm2StartupInit call
   // Initialize the TPM before loading the DXE core
   Status = Tpm2StartupInit ();
-  ASSERT_EFI_ERROR (Status);
```
kuqin12, line 189: MU_CHANGE, please?

kuqin12 (approved):

ArmPlatformPkg/PeilessSec/PeilessSec.c
Raymond-MS: It's between a MU_CHANGE [BEGIN] and a MU_CHANGE [END]. It should encapsulate this change as well. Plus, this wasn't really removed, it was just moved down below the if check now. :)

cfernald (approved):

